PR #2, opened 2017-05-23
Ad CRUD in ACP

senky:
WIP


adm/style/manage_ads.html
```diff
@@ -0,0 +1,36 @@
...
+			<th style="width: 80%;">{{ lang('AD_NAME') }}</th>
+			<th style="width: 10%;">{{ lang('AD_ENABLED') }}</th>
+			<th style="width: 10%;">{{ lang('ACTION') }}</th>
+		</tr>
+	</thead>
+	<tbody>
+		{% for ad in loops.ads %}
+			<tr>
+				<td><strong>{ad.NAME}</strong></td>
+				<td><a href="{ad.U_ENABLE}" title="{% if ad.ENABLED %}{{ lang('CLICK_TO_DISABLE') }}{% else %}{{ lang('CLICK_TO_ENABLE') }}{% endif %}">{% if ad.ENABLED %}{{ lang('ENABLED') }}{% else %}{{ lang('DISABLED') }}{% endif %}</a></td>
```
VSEphpbb: One thing you could try here which is kinda cool is make the lang vars an array:
```
'AD_ENABLE_TITLE' => array(
    0 => 'Click to enable',
    1 => 'Click to disable',
),
```
Then instead of the if/then logic for the title (and also works for the text in this line) just give the lang function the boolean switch:

```
title="{{ lang('AD_ENABLE_TITLE', ad.ENABLED) }}"
```

controller/admin_controller.php
```diff
@@ -0,0 +1,121 @@
...
+		$sql = 'SELECT ad_id, ad_name, ad_enabled
+			FROM ' . $this->ads_table;
+		$result = $this->db->sql_query($sql);
+		while ($row = $this->db->sql_fetchrow($result))
+		{
+			$ad_enabled = (bool) $row['ad_enabled'];
+
+			$this->template->assign_block_vars('ads', [
+				'NAME'		=> $row['ad_name'],
+				'ENABLED'	=> $ad_enabled,
```
VSEphpbb: This var should be named `S_ENABLED` since it is used as a boolean switch

language/en/info_acp_admanagement.php
```diff
@@ -20,4 +20,11 @@
 
 $lang = array_merge($lang, array(
 	'ACP_ADMANAGEMENT_TITLE'	=> 'Advertisement Management',
+
+	'AD_NAME'			=> 'Name',
```
VSEphpbb: I might go with calling this "Advertisement" instead of "Name" ?

VSEphpbb: info_acp_* files are loaded throughout the entire ACP. So this type of language file should only contain always needed lang vars, usually just lang vars used in the logs and in the titles of the modules. i.e.: ACP_ADMANAGEMENT_TITLE

create a separate stand-alone language file to hold lang keys that are only going to be used within the ad management ACP panel.

language/en/info_acp_admanagement.php
```diff
@@ -20,4 +20,11 @@
...
+	'AD_ENABLED'		=> 'Enabled',
+	'CLICK_TO_DISABLE'	=> 'Click to disable',
+	'CLICK_TO_ENABLE'	=> 'Click to enable',
+	'ACP_ADS_EMPTY'		=> 'No advertisement, yet. Add one with the button below.',
+	'ACP_ADS_ADD'		=> 'Add new advertisement',
```
VSEphpbb: I assume there will be options to edit and delete ads at some point too?

adm/style/manage_ads.html
senky: This is so cool! Isn't it possible, that such form will confuse translators? I will add a comment there, but still.

senky: I will retain text as it is, because it uses already existing language variables from `acp/common`, and I don't want to redefine them just to use this nice hack. Do you agree?

language/en/info_acp_admanagement.php
senky: Yes of course. Currently, it's only listing (you need to add entries manually into DB). Next goal is add, then edit, then delete and finally preview - for now only in one fixed template location, since template locations is planned later (see trello card).

adm/style/manage_ads.html
VSEphpbb: You can do the trick on the title but leave the ENABLED/DISABLED one's as is. And no it should not confuse translators. phpBB uses this technique a lot in the core language files.

language/en/acp.php
```diff
@@ -0,0 +1,30 @@
...
+}
+
+if (empty($lang) || !is_array($lang))
+{
+	$lang = array();
+}
+
+$lang = array_merge($lang, array(
+	'AD_NAME'			=> 'Name',
+	'AD_ENABLED'		=> 'Enabled',
```
VSEphpbb: This is not needed I think.

controller/admin_controller.php
```diff
@@ -0,0 +1,132 @@
+<?php
+/**
+*
+* Pages extension for the phpBB Forum Software package.
```
vinny: Pages

adm/style/manage_ads.html
```diff
@@ -0,0 +1,36 @@
...
+			<tr>
+				<td colspan="3" style="text-align: center;">{{ lang('ACP_ADS_EMPTY') }}</td>
+			</tr>
+		{% endfor %}
+	</tbody>
+</table>
+
+<form id="acp_admanagement" method="post" action="{{ U_ACTION }}">
+	<fieldset class="quick">
+		<input class="button2" type="submit" name="addpage" value="{{ lang('ACP_ADS_ADD') }}" />
```
VSEphpbb: name="addpage" should be changed (newad or something)

adm/style/manage_ads.html
```diff
@@ -0,0 +1,36 @@
...
+		<input type="hidden" name="action" value="add" />
```
VSEphpbb: I don't know if this line is needed. The "action" could already be part of the form's action, assuming it will have `?action=add` in it (you may need to define a template var for this).

controller/admin_controller.php
```diff
@@ -0,0 +1,132 @@
...
+*
+* @copyright (c) 2014 phpBB Limited <https://www.phpbb.com>
```
VSEphpbb: 2014 (pages is old 😉 )

controller/admin_controller.php
```diff
@@ -0,0 +1,132 @@
...
+
+	/**
+	* Constructor
+	*
+	* @param \phpbb\db\driver\driver_interface	$db			DB driver interface
+	* @param \phpbb\template\template			$template	Template object
+	* @param \phpbb\user						$user		User object
+	* @param \phpbb\request\request				$request	Request object
+	* @param string								$ads_table	Ads table
+	* @access public
```
VSEphpbb: we don't need `@access` if you don't want to bother with them in docblocks.

controller/admin_controller.php
```diff
@@ -0,0 +1,132 @@
...
+	* @return void
+	* @access public
+	*/
+	public function load_lang()
+	{
+		$this->user->add_lang_ext('phpbb/admanagement', 'acp');
+	}
+
+	/**
+	* Get geric ACP page title for Ads module
```
VSEphpbb: typo in "geric" but you can just lose this word all together.

adm/style/manage_ads.html
VSEphpbb: Noticed you have this var ready: `U_ADD_AD`

VSEphpbb: is `U_ADD_AD` which you defined meant to be used in this form's action?

controller/admin_controller.php
```diff
@@ -0,0 +1,132 @@
...
+				'U_DELETE'	=> $this->u_action . '&amp;action=delete&amp;id=' . $row['ad_id'], // TODO: ACP method
+			]);
+		}
+		$this->db->sql_freeresult($result);
+
+		// Set output vars for display in the template
+		$this->template->assign_vars(array(
+			'U_ACTION'	=> $this->u_action,
+			'U_ADD_AD'	=> $this->u_action . '&amp;action=add',
+		));
```
VSEphpbb: I think since you are modifying the u_action variable here, you can call these keys: `U_ACTION_ADD`, `U_ACTION_EDIT` and `U_ACTION_DELETE`

controller/admin_controller.php
senky: 😇

senky:
> So you're gradually going to add migrations when adding new functionality right? Database scheme looks good so far and doesn't need a UML diagram because it's still very simple.

@Derky I rather thought of not touching migrations after PR is merged, but modify them until then (instead of creating long set of migration files each updating just one column).

language/en/acp.php
```diff
@@ -19,11 +19,29 @@
...
+	'AD_NAME_REQUIRED'			=> 'Name is required.',
+	'AD_NAME_TOO_LONG'			=> 'Name length is limited to 255 characters.',
+	'ACP_AD_ADD_SUCCESS'		=> 'Advertisement added successfully!',
+	'ACP_AD_EDIT_SUCCESS'		=> 'Advertisement edited successfully!',
+	'ACP_AD_DELETE_SUCCESS'		=> 'Advertisement deleted successfully!',
+	'ACP_AD_DELETE_ERRORED'		=> 'There was an error deleting advertisement!',
+	'ACP_AD_ENABLE_SUCCESS'		=> 'Advertisement enabled successfully!',
+	'ACP_AD_DISABLE_SUCCESS'	=> 'Advertisement disabled successfully!',
+	'ACP_AD_ENABLE_ERRORED'		=> 'There was an error enabling advertisement!',
+	'ACP_AD_DISABLE_ERRORED'	=> 'There was an error disabling advertisement!',
```
VSEphpbb: Can you group them, the enables together and the disables together, like u did with delete above.

adm/style/manage_ads.html
```diff
@@ -1,36 +1,89 @@
...
+
+	<form id="acp_addmanagement_add" method="post" action="{{ U_ACTION }}">
+		<fieldset>
+			<dl>
+				<dt><label for="ad_name">{{ lang('AD_NAME') }}{{ lang('COLON') }}</label><br /><span>{{ lang('AD_NAME_EXPLAIN') }}</span></dt>
+				<dd><input class="text medium" type="text" id="ad_name" name="ad_name" value="{{ AD_NAME }}" maxlength="255" /></dd>
+			</dl>
+			<dl>
+				<dt><label for="ad_note">{{ lang('AD_NOTE') }}{{ lang('COLON') }}</label><br /><span>{{ lang('AD_NOTE_EXPLAIN') }}</span></dt>
+				<dd><textarea name="ad_note" rows="10" cols="60" style="width: 95%;">{{ AD_NOTE }}</textarea></dd>
```
VSEphpbb: these 2 text area's are missing id attributes

adm/style/scripts.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * This callback replaces enable links with disable links and vice versa.
+ * It does this by replacing the text, and replacing all instances of "enable"
+ * in the href with "disable", and vice versa.
+ */
+phpbb.addAjaxCallback('toggle_enable', function(res) {
+	var $this = $(this),
```
VSEphpbb: should have `'use strict'`

controller/admin_controller.php
```diff
@@ -89,19 +85,213 @@ public function get_page_title()
...
+			'AD_NOTE'		=> $data['ad_note'],
+			'AD_CODE'		=> $data['ad_code'],
+			'AD_ENABLED'	=> $data['ad_enabled'],
+		));
+	}
+
+	/**
+	* Enable/disable ad
+	*
+	* @return void
```
VSEphpbb: missing `@param`

controller/admin_controller.php
```diff
@@ -89,19 +85,213 @@ public function get_page_title()
...
+
+	/**
+	* Process 'edit' action
+	*
+	* @return void
+	*/
+	public function action_edit()
+	{
+		$ad_id = $this->request->variable('id', 0);
+		$data = $errors = array();
```
VSEphpbb: $data actually does not need to be set here. It is never used before being overwritten in some way by your cases below.

controller/admin_controller.php
```diff
@@ -89,19 +85,213 @@ public function get_page_title()
...
+		}
+		else
+		{
+			$sql = 'SELECT *
+				FROM ' . $this->ads_table . '
+				WHERE ad_id = ' . (int) $ad_id;
+			$result = $this->db->sql_query($sql);
+			$rowset = $this->db->sql_fetchrowset($result);
+			$data = $rowset[0];
+			$this->db->sql_freeresult($result);
```
VSEphpbb: Should maybe handle an error in case this query has no results/errors.

controller/admin_controller.php
```diff
@@ -0,0 +1,322 @@
...
+	}
+
+	/**
+	* Process 'delete' action
+	*
+	* @return void
+	*/
+	public function action_delete()
+	{
+		$sql = 'DELETE FROM ' . $this->ads_table . '
```
VSEphpbb: There should be a confirm_box to prevent accidental deletion.

VSEphpbb:
Now that I see it in action in the ACP, I think the module mode should be "manage" instead of settings, and the title for it "Manage ads"

adm/style/manage_ads.html
```diff
@@ -1,36 +1,89 @@
...
+		<div class="errorbox">
+			<h3>{{ lang('WARNING') }}</h3>
+			<p>{{ ERROR_MSG }}</p>
+		</div>
+	{% endif %}
+
+	<form id="acp_addmanagement_add" method="post" action="{{ U_ACTION }}">
+		<fieldset>
+			<dl>
+				<dt><label for="ad_name">{{ lang('AD_NAME') }}{{ lang('COLON') }}</label><br /><span>{{ lang('AD_NAME_EXPLAIN') }}</span></dt>
```
senky: 🆒

language/en/acp.php
```diff
@@ -19,11 +19,29 @@
...
+	'AD_ENABLED_EXPLAIN'	=> 'If disabled, this advertisement will not be displayed to this board users.',
+	'AD_ENABLE_TITLE'		=> array( // Plural rule doesn't apply here! Just translate the values.
 		0 => 'Click to enable',
 		1 => 'Click to disable',
 	),
 	'ACP_ADS_EMPTY'		=> 'No advertisement, yet. Add one with the button below.',
 	'ACP_ADS_ADD'		=> 'Add new advertisement',
+
+	'AD_NAME_REQUIRED'			=> 'Name is required.',
+	'AD_NAME_TOO_LONG'			=> 'Name length is limited to 255 characters.',
```
senky: I don't like this. It's hardcoded in migration, too. To create really bulletproof solution, migration file would have to access to the constant as well. Furthermore, why would we want to change the max length? If in any case, then enlarge it to `TEXT` without limits.

VSEphpbb: It's in the migration because that is the physical limitation of a VARCHAR field in most of the DBMS. So it is fine there and it should always be 255 even if we wanted to make it smaller in practice.

But in this case here it is a magic number. And if we decided to make it 100 chars, we'd never change the DB set-up, we just check the length from user input here. We would only ever want to make it smaller, but the issue isn't whther we'd change it or not so much as it's a magic number, and we do not allow magic numbers in coding guidelines:

> "Magic Numbers: Don't use them. Use named constants for any literal value other than obvious special cases."

adm/style/scripts.js
```diff
@@ -1,3 +1,5 @@
+'use strict'
+
```
VSEphpbb: This should be the first line inside your function before var $this...

controller/admin_controller.php
```diff
@@ -214,9 +214,13 @@ public function action_edit()
 				FROM ' . $this->ads_table . '
 				WHERE ad_id = ' . (int) $ad_id;
 			$result = $this->db->sql_query($sql);
-			$rowset = $this->db->sql_fetchrowset($result);
-			$data = $rowset[0];
+			$data = $this->db->sql_fetchrow($result);
 			$this->db->sql_freeresult($result);
+
+			if (!$data)
```
VSEphpbb: I would use empty($data) since you could possibly get either a "false" or "array()" as values.

controller/admin_controller.php
```diff
@@ -267,23 +272,44 @@ public function ad_enable($enable)
...
-		$this->db->sql_query($sql);
+		$ad_id = $this->request->variable('id', 0);
 
-		if ($this->db->sql_affectedrows())
-		{
-			trigger_error($this->user->lang('ACP_AD_DELETE_SUCCESS') . adm_back_link($this->u_action));
-		}
-		else
+		$sql = 'SELECT ad_id
+			FROM ' . $this->ads_table . '
```
VSEphpbb: This query seems unnecessary.

controller/admin_controller.php
```diff
@@ -267,23 +272,44 @@ public function ad_enable($enable)
...
+				if (!$this->db->sql_affectedrows())
+				{
+					trigger_error($this->user->lang('ACP_AD_DELETE_ERRORED') . adm_back_link($this->u_action), E_USER_WARNING);
+				}
+			}
+			else
+			{
+				confirm_box(false, $this->user->lang('CONFIRM_OPERATION'), build_hidden_fields(array(
+					'id'		=> $ad_id,
+					'i'			=> $this->request->variable('i', ''),
```
VSEphpbb: what is 'i'?

controller/admin_controller.php
```diff
@@ -267,23 +272,44 @@ public function ad_enable($enable)
...
-		}
-		else
+		$sql = 'SELECT ad_id
+			FROM ' . $this->ads_table . '
+			WHERE ad_id = ' . (int) $ad_id;
+		$result = $this->db->sql_query($sql);
+		$row = $this->db->sql_fetchrow($result);
+		$this->db->sql_freeresult($result);
+
+		if ($row)
```
VSEphpbb: instead of `if ($row)` you could lose the query above and do`if ($ad_id)`

acp/main_info.php
```diff
@@ -21,7 +21,7 @@ public function module()
 			'filename'	=> '\phpbb\admanagement\acp\main_module',
 			'title'		=> 'ACP_ADMANAGEMENT_TITLE',
 			'modes'		=> array(
-				'settings'	=> array(
+				'manage'	=> array(
 					'title'	=> 'ACP_ADMANAGEMENT_TITLE',
```
VSEphpbb: This title should also be updated to a new key for "Manage ads"

controller/admin_controller.php
senky: Query is meant to ensure admin is not confronted with the confirm box in case ad does not exist. But I have removed it for now and user is given warning after he confirms deletion.

VSEphpbb: It's still too costly/excessive for that. If it does not truly exist then nothing will be deleted, so it is still better not to have this

VSEphpbb: IMO this would only be worthwhile if you were gonna actually use, for example to put the name of the ad into the confirm box message: Are you sure you want to delete "Some Ad Name"?

controller/admin_controller.php
```diff
@@ -322,20 +327,21 @@ public function list_ads()
...
+				'U_EDIT'	=> $this->u_action . '&amp;action=edit&amp;id=' . $row['ad_id'],
+				'U_DELETE'	=> $this->u_action . '&amp;action=delete&amp;id=' . $row['ad_id'],
 			));
 		}
 		$this->db->sql_freeresult($result);
 
 		// Set output vars for display in the template
 		$this->template->assign_vars(array(
 			'U_ACTION_ADD'	=> $this->u_action . '&amp;action=add',
+			'ICON_PREVIEW'	=> '<img src="' . htmlspecialchars($phpbb_admin_path) . 'images/file_up_to_date.gif" alt="' . $this->user->lang('AD_PREVIEW') . '" title="' . $this->user->lang('AD_PREVIEW') . '" />',
```
VSEphpbb: $phpbb_admin_path is not defined.

VSEphpbb: The icon doesn't seem right. Maybe come up with our own?

controller/admin_controller.php
```diff
@@ -322,20 +327,21 @@ public function list_ads()
...
-			$this->template->assign_block_vars('ads', array( // TODO: convert back to original notation (3.1 does not support this)
+			$this->template->assign_block_vars('ads', array(
 				'NAME'		=> $row['ad_name'],
 				'S_ENABLED'	=> (int) $ad_enabled,
-				'U_ENABLE'	=> $this->u_action . '&amp;action=' . ($ad_enabled ? 'disable' : 'enable') . '&amp;id=' . $row['ad_id'], // TODO: ACP method
-				'U_PREVIEW'	=> '', // TODO: frontend logic
-				'U_EDIT'	=> $this->u_action . '&amp;action=edit&amp;id=' . $row['ad_id'], // TODO: ACP method
-				'U_DELETE'	=> $this->u_action . '&amp;action=delete&amp;id=' . $row['ad_id'], // TODO: ACP method
+				'U_ENABLE'	=> $this->u_action . '&amp;action=' . ($ad_enabled ? 'disable' : 'enable') . '&amp;id=' . $row['ad_id'],
+				'U_PREVIEW'	=> append_sid(generate_board_url() . '/index.' . $this->php_ext, 'ad_preview=' . $row['ad_id']),
```
VSEphpbb: I don't think this should say `ad_preview`. In my browser, my ad content blocker actually notices this and prevents me from previewing it. Using `preview_ad` does not trigger my content blocker. But I'm open to other better names for this action too.

event/main_listener.php
```diff
@@ -20,26 +20,64 @@
...
-		$forum_row = $event['forum_row'];
-		$forum_row['FORUM_NAME'] .= ' :: Acme Event ::';
-		$event['forum_row'] = $forum_row;
+		$ad_preview = $this->request->variable('ad_preview', 0);
+
+		$sql = 'SELECT ad_code
+			FROM ' . $this->ads_table . '
+			WHERE ad_id = ' . (int) $ad_preview;
+		$result = $this->db->sql_query($sql);
+		$row = $this->db->sql_fetchrow($result);
```
VSEphpbb: How about just `sql_fetchfield` since we're only grabbing one thing (unless you plan on  expanding this query later?)

controller/admin_controller.php
```diff
@@ -349,4 +349,20 @@ public function list_ads()
 			'ICON_PREVIEW'	=> '<img src="' . htmlspecialchars($this->phpbb_admin_path) . 'images/file_up_to_date.gif" alt="' . $this->user->lang('AD_PREVIEW') . '" title="' . $this->user->lang('AD_PREVIEW') . '" />',
 		));
 	}
+
+	/**
+	* @param	mixed	...$lang	Input to $user->lang()
```
VSEphpbb: These should have titles in their docblocks too. Also @param is not needed. But you could instead write in the description that it takes arguments in the form of a Llanguage key, followed by language substitution values.

controller/admin_controller.php
```diff
@@ -0,0 +1,368 @@
...
+					$this->error('ACP_AD_DELETE_ERRORED');
+				}
+			}
+			else
+			{
+				confirm_box(false, $this->user->lang('CONFIRM_OPERATION'), build_hidden_fields(array(
+					'id'		=> $ad_id,
+					'i'			=> $this->request->variable('i', ''),
+					'mode'		=> $this->request->variable('mode', ''),
+					'action'	=> 'delete'))
```
VSEphpbb: these 2 trailing parentheses should be on the line below with that lonely parenthesis.

controller/admin_controller.php
```diff
@@ -0,0 +1,368 @@
...
+			// Validate data
+			if ($data['ad_name'] === '')
+			{
+				$errors[] = $this->user->lang('AD_NAME_REQUIRED');
+			}
+			if (truncate_string($data['ad_name'], self::MAX_NAME_LENGTH) !== $data['ad_name'])
+			{
+				$errors[] = $this->user->lang('AD_NAME_TOO_LONG', self::MAX_NAME_LENGTH);
+			}
+
```
VSEphpbb: The top half of this function is an exact duplicate of the action_add function. 

Since we want to avoid code duplication you should extract the duplicated code into its own separate (protected) methods.

I see we can have the following here in both the add_action and edit_action functions:
```php
$this->check_form_key();

$data = $this->get_form_data();

$this->validate($data['ad_name']);
```
Of course `$errors` whould need to be made a class property.

VSEphpbb:
!set WIP

VSEphpbb:
By the way @senky you can type `!set WIP` or `!unset WIP` to add or remove the work in progress label on your PRs. That way you can let it be known to us. You can actually do this for any of our labels (ready for review, needs tests, bug, etc..).

senky:
!set help wanted

senky:
@VSEphpbb Tests added. Some of them fail, however. See `TODO`s in test files. Can you help me with them? Those are the parts that I really have no idea how to fix.

tests/functional/acp_test.php
```diff
@@ -0,0 +1,210 @@
...
+
+		// Confirm error when submitting without required field data
+		$form = $crawler->selectButton($this->lang('SUBMIT'))->form();
+		$crawler = self::submit($form);
+		$this->assertGreaterThan(0, $crawler->filter('.errorbox')->count());
+		$this->assertContainsLang('AD_NAME_REQUIRED', $crawler->text());
+
+		// Confirm error when submitting too long ad name
+		$form_data = array(
+			'ad_name'		=> 'Lorem ipsum dolor sit amet, consectetuer adipiscing elit. Aenean commodo ligula eget dolor. Aenean massa. Cum sociis natoque penatibus et magnis dis parturient montes, nascetur ridiculus mus. Donec quam felis, ultricies nec, pellentesque eu, pretium quis, sem. Nulla consequat massa quis enim. Donec.',
```
VSEphpbb: is simpler just to use `str_repeat('a', 256)` instead of an actual string.

tests/functional/acp_test.php
```diff
@@ -0,0 +1,210 @@
...
+		$form = $crawler->selectButton($this->lang('SUBMIT'))->form();
+		$crawler = self::submit($form, $form_data);
+		$this->assertGreaterThan(0, $crawler->filter('.errorbox')->count());
+		$this->assertContains($this->lang('AD_NAME_TOO_LONG', 255), $crawler->text());
+
+		// Create ad
+		$form_data = array(
+			'ad_name'		=> 'Functional test name',
+			'ad_note'		=> 'Functional test note',
+			'ad_code'		=> '<img src="https://www.phpbb.com/assets/images/images/logo_phpbb.png" />',
```
VSEphpbb: Can you use something less reliant on something that may or may not be able to exist?
``` `` for example?

File: tests/controller/admin_controller_test.php [hunk 36]
```diff
@@ -0,0 +1,319 @@
...
+				WHERE ad_name = "' . $ad_name . '"';
+			$result = $this->db->sql_query($sql);
+			$row = $this->db->sql_fetchrow($result);
+
+			$this->assertEquals('', $row['ad_note']);
+			$this->assertEquals('', $row['ad_code']);
+			$this->assertEquals(0, $row['ad_enabled']);
+		}
+
+		// TODO: fails because trigger_error is called
```
VSEphpbb: Use `setExpectedTriggerError`. It's like the mock->expects() in that it must be set before you actually run the code that will call trigger_error.

Here's some examples:
https://github.com/phpbb/ideas/blob/master/tests/controller/idea_controller_test.php#L92
https://github.com/phpbb-extensions/autogroups/blob/master/tests/controller/submit_autogroup_rule_test.php#L98

tests/controller/admin_controller_test.php
```diff
@@ -0,0 +1,319 @@
...
+		if ($ad_id)
+		{
+			$sql = 'SELECT ad_enabled
+				FROM ' . $this->ads_table . '
+				WHERE ad_id = ' . $ad_id;
+			$result = $this->db->sql_query($sql);
+			$ad_enabled = (bool) $this->db->sql_fetchfield('ad_enabled', $result);
+			$this->db->sql_freeresult($result);
+
+			$this->assertEquals(!$enabled, $ad_enabled);
```
VSEphpbb: `!$enable`

tests/controller/admin_controller_test.php
```diff
@@ -0,0 +1,319 @@
...
+ * Advertisement management. An extension for the phpBB Forum Software package.
+ *
+ * @copyright (c) 2017 phpBB Limited <https://www.phpbb.com>
+ * @license GNU General Public License, version 2 (GPL-2.0)
+ *
+ */
+
+namespace phpbb\admanagement\tests\controller;
+
+require_once dirname(__FILE__) . '/../../../../../includes/functions.php';
```
VSEphpbb: do not need this

tests/controller/admin_controller_test.php
```diff
@@ -0,0 +1,319 @@
...
+require_once dirname(__FILE__) . '/../../../../../includes/functions_content.php';
```
VSEphpbb: do not need this

tests/controller/admin_controller_test.php
```diff
@@ -0,0 +1,319 @@
...
+require_once dirname(__FILE__) . '/../../../../../includes/functions_acp.php';
```
VSEphpbb: use __DIR__ instead of dirname()

tests/controller/admin_controller_test.php
```diff
@@ -0,0 +1,319 @@
...
+	/**
+	* Test action_delete() method
+	*/
+	public function test_action_delete()
+	{
+		$controller = $this->get_controller();
+
+		$this->request->method('variable')->willReturn(1);
+
+		// TODO: how to set `confirm_box(true) == true`?
```
VSEphpbb: Here you can see how to mock confirm_box, add_form_key and check_form_key:
https://github.com/phpbb-extensions/topicprefixes/blob/master/tests/controller/admin_controller_base.php#L90-L112

Note that for them to work, the namespace of the test must actually be the same namespace as the admin_controller
https://github.com/phpbb-extensions/topicprefixes/blob/master/tests/controller/admin_controller_base.php#L15

Also look at how we used static vars with for these values in these tests:
https://github.com/phpbb-extensions/topicprefixes/blob/master/tests/controller/admin_controller_base.php#L22-L25

VSEphpbb: So here's an example test with trigger_errors and confirm boxes.
https://github.com/phpbb-extensions/topicprefixes/blob/master/tests/controller/delete_prefix_test.php

tests/controller/admin_controller_test.php
VSEphpbb: example: https://github.com/phpbb-extensions/topicprefixes/blob/master/tests/controller/admin_controller_base.php#L17

tests/controller/admin_controller_test.php
```diff
@@ -0,0 +1,350 @@
...
+
+	/**
+	* Test data for the test_action_add_submit() function
+	*
+	* @return array Array of test data
+	*/
+	public function action_add_data()
+	{
+		return array(
+			array('', true, 'Name is required.'),
```
VSEphpbb: Use lang keys here instead

tests/controller/admin_controller_test.php
```diff
@@ -0,0 +1,350 @@
...
+				FROM ' . $this->ads_table . '
+				WHERE ad_id = ' . $ad_id;
+			$result = $this->db->sql_query($sql);
+			$ad_enabled = (bool) $this->db->sql_fetchfield('ad_enabled', $result);
+			$this->db->sql_freeresult($result);
+
+			$this->assertEquals(!$enable, $ad_enabled);
+		}
+
+		//  should check for trigger_error lang here
```
VSEphpbb: is this a todo?

tests/controller/admin_controller_test.php
```diff
@@ -0,0 +1,350 @@
...
+	public function test_action_add_submit($ad_name, $s_error, $error_msg)
+	{
+		$controller = $this->get_controller();
+
+		$this->request->expects($this->once())
+			->method('is_set_post')
+			->with('submit')
+			->willReturn(true);
+
+		$this->request->method('variable')->will($this->onConsecutiveCalls($ad_name, '', '', false));
```
VSEphpbb: each chained method call should be on a new line.

tests/controller/admin_controller_test.php
```diff
@@ -0,0 +1,350 @@
...
+					'ERROR_MSG'	=> $error_msg,
+				));
+
+			$this->template->expects($this->at(1))
+				->method('assign_vars')
+				->with(array(
+					'AD_NAME'		=> $ad_name,
+					'AD_NOTE'		=> '',
+					'AD_CODE'		=> '',
+					'AD_ENABLED'	=> false,
```
VSEphpbb: why is this false here, but 0 later on below?

I would set up a data array for these values at the top of this function, and just use the array instead of "raw" values in the places where you are testing them...

i.e.:
```
$data = array(
    'ad_name' => $ad_name,
    'ad_note' => '',
    'ad_code' => '',
    'ad_enabled' => 0,
);
```
Then just use `$data['ad_note']` instead of `''` for example.

tests/controller/admin_controller_test.php
senky: Well, since it's probably impossible to test trigger_error lang, I will just remove it.

VSEphpbb: You can test what the message is in setExpectedTrigger.

It takes the message as it's 2nd arg.

tests/controller/admin_controller_test.php
senky: Template assign_vars really returns `false` (`$this->request->variable('ad_enabled', false)`), while data from DB return `'0'`. So I will replace `$this->assertEquals(0, $row['ad_enabled']);` then.

VSEphpbb: Ah, well that is because you do this in the admin_controller
```
$ad_enabled = (bool) $row['ad_enabled'];
```
I would in the controller simply cast it to (int) all the time...so it is always 0. The template will recognize 0 as a false.

event/main_listener.php
```diff
@@ -20,26 +20,64 @@
...
+			FROM ' . $this->ads_table . '
+			WHERE ad_id = ' . (int) $ad_preview;
+		$result = $this->db->sql_query($sql);
+		$ad_code = $this->db->sql_fetchfield('ad_code', $result);
+		$this->db->sql_freeresult($result);
+
+		if (!empty($ad_code))
+		{
+			$this->template->assign_vars(array(
+				'AD_PREVIEW'	=> true,
```
VSEphpbb: boolean switch should be S_AD_PREVIEW

event/main_listener.php
```diff
@@ -20,26 +20,64 @@
...
 	 */
-	public function display_forums_modify_template_vars($event)
+	public function ad_preview()
 	{
-		$forum_row = $event['forum_row'];
-		$forum_row['FORUM_NAME'] .= ' :: Acme Event ::';
-		$event['forum_row'] = $forum_row;
+		$ad_preview = $this->request->variable('ad_preview', 0);
+
+		$sql = 'SELECT ad_code
```
VSEphpbb: Need to return if there is no ad_preview. Currently this is always running this query.

event/main_listener.php
```diff
@@ -20,26 +20,64 @@
...
+			FROM ' . $this->ads_table . '
+			WHERE ad_id = ' . (int) $ad_preview;
+		$result = $this->db->sql_query($sql);
+		$ad_code = $this->db->sql_fetchfield('ad_code', $result);
```
VSEphpbb: Remove $result... Only use 'ad_code'

tests/controller/admin_controller_test.php
```diff
@@ -0,0 +1,377 @@
...
+		$controller->action_delete();
+
+		$sql = 'SELECT ad_id
+			FROM ' . $this->ads_table . '
+			WHERE ad_id = 1';
+		$result = $this->db->sql_query($sql);
+		$row = $this->db->sql_fetchrow($result);
+		$this->db->sql_freeresult($result);
+
+		$this->assertEquals(true, empty($row));
```
VSEphpbb: `$this->assertTrue(empty($row));`

tests/controller/admin_controller_test.php
```diff
@@ -0,0 +1,377 @@
...
+	}
+
+	/**
+	* Test action_delete() method
+	*/
+	public function test_action_delete()
+	{
+		$controller = $this->get_controller();
+
+		$this->request->method('variable')->willReturn(1);
```
VSEphpbb: separate lines please

VSEphpbb: should have an `expects($this->any())` call before the `method()` (here and anywhere else)

tests/controller/admin_controller_test.php
```diff
@@ -0,0 +1,377 @@
...
+	}
+
+	/**
+	* Test list_ads() method
+	*/
+	public function test_list_ads()
+	{
+		$controller = $this->get_controller();
+
+		$this->template->expects($this->atLeastOnce())->method('assign_block_vars');
```
VSEphpbb: lines

tests/controller/admin_controller_test.php
```diff
@@ -0,0 +1,377 @@
...
+ * Mock confirm_box()
+ * Note: use the same namespace as the admin_controller
+ *
+ * @return bool
+ */
+function confirm_box()
+{
+	return \phpbb\admanagement\controller\admin_controller_test::$confirm;
+}
+/**
```
VSEphpbb: an empty line between methods please

tests/controller/fixtures/ad.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<dataset>
```
VSEphpbb: this file mixed spaces and tabs

tests/event/fixtures/ad.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<dataset>
```
VSEphpbb: As you grow this extension it will become a pain to maintain multiple test fixtures. You can just make one fixture and keep a fixtures folder in the root of the tests folder. Then all tests can reference just the single test fixture.

tests/event/main_listener_base.php
```diff
@@ -0,0 +1,71 @@
...
+
+	/**
+	* {@intheritDoc}
+	*/
+	public function setUp()
+	{
+		parent::setUp();
+
+		// Load/Mock classes required by the listener class
+        $this->request = $this->getMock('\phpbb\request\request');
```
VSEphpbb: mixed tabs and spaces

tests/event/main_listener_test.php
```diff
@@ -0,0 +1,33 @@
...
+	/**
+	* Test the event listener is subscribing events
+	*/
+	public function test_getSubscribedEvents()
+	{
+		$this->assertEquals(array(
+			'core.page_header_after',
+		), array_keys(\phpbb\admanagement\event\main_listener::getSubscribedEvents()));
+	}
+}
```
VSEphpbb: missing new line at end of file.

controller/admin_controller.php
```diff
@@ -0,0 +1,431 @@
...
+				$sql = 'UPDATE ' . $this->ads_table . '
+					SET ' . $this->db->sql_build_array('UPDATE', $data) . '
+					WHERE ad_id = ' . (int) $ad_id;
+				$this->db->sql_query($sql);
+
+				$this->success('ACP_AD_EDIT_SUCCESS');
+			}
+			else
+			{
+				$this->assign_errors();
```
VSEphpbb: It looks like this is not needed. What I mean is, assign_errors() and assign_form_data() can be merged into one function. assign_errors is used immediately before assign_form_data up above. But as for here, well, either we submitted with success so the trigger_error means we never get this far...or there either were errors or we were not submitting which means assign_form_data at the end of this function would be able to assign any errors if there are any at that point.

controller/admin_controller.php
```diff
@@ -0,0 +1,431 @@
...
+
+			if (empty($this->errors))
+			{
+				// Insert the ad data to the database
+				$sql = 'INSERT INTO ' . $this->ads_table . ' ' . $this->db->sql_build_array('INSERT', $data);
+				$this->db->sql_query($sql);
+
+				$this->success('ACP_AD_ADD_SUCCESS');
+			}
+			else
```
VSEphpbb: This else statement isn't really needed. If there are no errors, we end execution at the trigger_error success message.

controller/admin_controller.php
```diff
@@ -0,0 +1,431 @@
...
+	*
+	* @return void
+	*/
+	public function main()
+	{
+		$this->user->add_lang_ext('phpbb/admanagement', 'acp');
+
+		switch ($this->request->variable('action', ''))
+		{
+			case 'add':
```
VSEphpbb: You could also try this if you like it:
```
case 'add':
case 'edit':
case 'delete':
    $this->{'action_' . $action}();
break;
```

controller/admin_controller.php
```diff
@@ -0,0 +1,431 @@
...
+	*
+	* @return	array	Form data
+	*/
+	protected function get_form_data()
+	{
+		return array(
+			'ad_name'		=> $this->request->variable('ad_name', '', true),
+			'ad_note'		=> $this->request->variable('ad_note', '', true),
+			'ad_code'		=> $this->request->variable('ad_code', '', true),
+			'ad_enabled'	=> $this->request->variable('ad_enabled', false),
```
VSEphpbb: ad_enabled in the database is an INT, default to 0. It should be the same here, using 0 instead of false

controller/admin_controller.php
```diff
@@ -0,0 +1,431 @@
...
+				'U_EDIT'	=> $this->u_action . '&amp;action=edit&amp;id=' . $row['ad_id'],
+				'U_DELETE'	=> $this->u_action . '&amp;action=delete&amp;id=' . $row['ad_id'],
+			));
+		}
+		$this->db->sql_freeresult($result);
+
+		// Set output vars for display in the template
+		$this->template->assign_vars(array(
+			'U_ACTION_ADD'	=> $this->u_action . '&amp;action=add',
+			'ICON_PREVIEW'	=> '<img src="' . htmlspecialchars($this->phpbb_admin_path) . 'images/file_up_to_date.gif" alt="' . $this->user->lang('AD_PREVIEW') . '" title="' . $this->user->lang('AD_PREVIEW') . '" />',
```
Derky: Please remove `htmlspecialchars()` here.

VSEphpbb: Here is an SVG 
[eye-open.svg.zip](https://github.com/phpbb-extensions/ad-management/files/1035394/eye-open.svg.zip)

VSEphpbb: Here's a PNG of a font awesome icon we can use in an ACP matched color 
eye-open

controller/admin_controller.php
senky: It's BOOL in the database.

senky:
!unset help wanted

senky:
!unset missing tests

controller/admin_controller.php
```diff
@@ -329,11 +285,46 @@ public function list_ads()
...
+	* Enable/disable an advertisement
+	*
+	* @param	bool	$enable	Enable or disable the advertisement?
+	* @return void
+	*/
+	protected function ad_enable($enable)
+	{
+		$sql = 'UPDATE ' . $this->ads_table . '
+			SET ad_enabled = ' . (int) $enable . '
+			WHERE ad_id = ' . (int) $this->request->variable('id', 0);
```
Derky: I prefer to put the`id` in a variable on a separate line before the query.

tests/controller/admin_controller_test.php
```diff
@@ -307,26 +453,70 @@ public function test_ad_enable($ad_id, $enable, $err_msg)
...
-			FROM ' . $this->ads_table . '
-			WHERE ad_id = 1';
-		$result = $this->db->sql_query($sql);
-		$row = $this->db->sql_fetchrow($result);
-		$this->db->sql_freeresult($result);
+		if ($confirm)
+		{
+			$sql = 'SELECT ad_id
+				FROM ' . $this->ads_table . '
+				WHERE ad_id = ' . $ad_id;
```
Derky: Please (int) cast `ad_id`.

tests/controller/admin_controller_test.php
```diff
@@ -261,6 +261,145 @@ public function test_action_add_submit($ad_name, $s_error, $error_msg)
...
+			$this->setExpectedTriggerError(E_USER_WARNING, 'ACP_AD_DOES_NOT_EXIST');
+		}
+
+		$controller->action_edit();
+
+		// Check ad is in the DB
+		if (!$s_error)
+		{
+			$sql = 'SELECT * FROM ' . $this->ads_table . '
+				WHERE ad_id = "' . $ad_id . '"';
```
Derky: Please also cast the id's with (int) in tests, it keeps catching my attention. 😛

controller/admin_controller.php
VSEphpbb: Well the thing is it should be treated with consistency throughout.

controller/admin_controller.php
```diff
@@ -329,11 +285,48 @@ public function list_ads()
 		// Set output vars for display in the template
 		$this->template->assign_vars(array(
 			'U_ACTION_ADD'	=> $this->u_action . '&amp;action=add',
-			'ICON_PREVIEW'	=> '<img src="' . htmlspecialchars($this->phpbb_admin_path) . 'images/file_up_to_date.gif" alt="' . $this->user->lang('AD_PREVIEW') . '" title="' . $this->user->lang('AD_PREVIEW') . '" />',
+			'ICON_PREVIEW'	=> '<img src="' . $this->ext_path . 'adm/images/icon_preview.png" alt="' . $this->user->lang('AD_PREVIEW') . '" title="' . $this->user->lang('AD_PREVIEW') . '" />',
```
VSEphpbb: One thing about the icon. The reason I provided a larger 32px icon is for retina displays. Basically you use an icon that is 2x as large as you want, then use CSS (either inline style or linked CSS) to scale the image to the dimensions you want it displayed. This makes the images crystal clear on both normal displays and 4k/5k/retina displays.

controller/admin_controller.php
```diff
@@ -402,6 +382,9 @@ protected function assign_errors()
 	protected function assign_form_data($data)
 	{
 		$this->template->assign_vars(array(
+			'S_ERROR'			=> (bool) count($this->errors),
```
VSEphpbb: Can these line up with the one's below?

controller/admin_controller.php
VSEphpbb: The actual value stored is either 1 or 0. We're always using it as an INT which is why it should be better to use 0 here istead of false.

senky: So you mean always use 0 and 1 instead of false and true? In templates, request::variable, SQL queries; simply everywhere?

controller/admin_controller.php
senky: I though of that, but since all other icons aren't retina-ready, why bother? If you really want, I will rewrite it.

controller/admin_controller.php
VSEphpbb: No I just mean set this to 0 instead of false here on this line. Everything else looks good regarding this variable.

controller/admin_controller.php
VSEphpbb: Good point. Don't bother this is fine.

PR #4, opened 2017-05-30
Template locations

senky:


controller/admin_controller.php
```diff
@@ -162,6 +192,26 @@ public function action_edit()
...
+				// Insert data to ad_locations table
+				$sql_ary = array();
+				foreach ($ad_locations as $ad_location)
+				{
+					$sql_ary[] = array(
+						'ad_id'			=> $ad_id,
+						'location_id'	=> $ad_location,
+					);
+				}
+				$this->db->sql_multi_insert($this->ad_locations_table, $sql_ary);
```
VSEphpbb: This insertion is the same as the one in the add_action. So I see an opportunity for some methods here. Maybe there can be `insert_ad_locations` and `delete_ad_locations` and `get_ad_locations` methods.

In the bigger picture, I think it's time now for a dedicated ad manager class? A class that can handle the loading, inserting, updating and deleting of ads and ad locations from the db. That would remove a lot of protected methods from this class and improve the score.

I think we'd need:
```php
public function get_ad($ad_id); // get a single ad

public function get_ads($ad_locations); // get ads at given link locations as done in event listener

public function get_all_ads(); // get every ad exists as done in ACP

public function insert_ad($data);

public function update_ad($ad_id, $data); // this would handle both action_edit and ad_enable updates

public function delete_ad($ad_id);

public function get_ad_locations($ad_id);

public function insert_ad_locations($ad_id, $ad_locations);

public function delete_ad_locations($ad_id);
```

event/main_listener.php
```diff
@@ -82,5 +92,24 @@ public function ad_preview()
 				));
 			}
 		}
+		else
+		{
+			$location_ids = $this->location_manager->get_all_location_ids();
+
+			$sql = 'SELECT al.location_id, a.ad_code
```
VSEphpbb: I don't usually like to see a lot of queries and logic inside an event listener. So I think it may be another reason here to have a dedicated ad manager. You could use a `get_ad($id)` method in the above preview code. And here you could have a `get_ads($locations)` sort of method.

location/manager.php
```diff
@@ -0,0 +1,83 @@
...
+		$this->register_template_locations($template_locations);
+	}
+
+	/**
+	* Get a list of all template location types
+	*
+	* Returns an associated array where key is the location id
+	* and value is array of location name and location description.
+	*
+	* @return	array	Array containing a list of all avatar drivers
```
VSEphpbb: avatar drivers?

location/type/type_interface.php
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+* Interface for template location types
+*/
+interface type_interface
+{
+	/**
+	* Returns the unique ID of the location.
+	*
+	* @return string	ID of location.
```
VSEphpbb: `_id` keeps making me thing this is about numbers. what do you think of renaming this as `_tag` everywhere instead?

event/main_listener.php
```diff
@@ -82,5 +92,24 @@ public function ad_preview()
...
+				LEFT JOIN ' . $this->ads_table . ' a
+					ON (al.ad_id = a.ad_id)
+				WHERE a.ad_enabled = 1
+					AND ' . $this->db->sql_in_set('al.location_id', $location_ids) . '
+				GROUP BY al.location_id';
+			$result = $this->db->sql_query($sql);
+			while ($row = $this->db->sql_fetchrow($result))
+			{
+				$this->template->assign_vars(array(
+					'AD_' . strtoupper($row['location_id'])	=> $row['ad_code'],
```
VSEphpbb: need to use htmlspecialchars_decode on the ad_code.

adm/style/manage_ads.html
```diff
@@ -34,6 +34,15 @@
 				<dd><label><input type="radio" class="radio" id="ad_enabled" name="ad_enabled" value="1"{% if AD_ENABLED %} checked="checked"{% endif %} /> {{ lang('ENABLED') }}</label>
 					<label><input type="radio" class="radio" name="ad_enabled" value="0"{% if not AD_ENABLED %} checked="checked"{% endif %} /> {{ lang('DISABLED') }}</label></dd>
 			</dl>
+			<dl>
+				<dt><label for="ad_locations">{{ lang('AD_LOCATIONS') ~ lang('COLON') }}</label><br /><span>{{ lang('AD_LOCATIONS_EXPLAIN') }}</span></dt>
```
VSEphpbb: The new language keys added here need to be added to the language file.

styles/prosilver/template/event/overall_header_content_before.html
```diff
@@ -1,3 +1,7 @@
 {% if S_AD_PREVIEW %}
     <div style="text-align: center;">{{ AD_CODE }}</div>
 {% endif %}
+
```
VSEphpbb: I think this should be in the event `overall_header_navbar_before`. This is currently placing this location below the nav-bar. This template event can be used for a new ad location named `AD_BELOW_NAVBAR`, for example.

location/type/type_interface.php
senky: It itches me as well, but `id` stands for `identifier`, what it really is. `tag` evokes in me something that you can assign to multiple entities to group them.

event/main_listener.php
```diff
@@ -54,18 +51,16 @@ static public function getSubscribedEvents()
...
+	* @param \phpbb\admanagement\ad\manager			$manager			Advertisement manager object
 	* @param \phpbb\admanagement\location\manager	$location_manager	Template location manager object
-	* @param string									$ads_table			Ads table
-	* @param string									$ad_locations_table	Ad locations table
 	*/
-	public function __construct(\phpbb\request\request $request, \phpbb\db\driver\driver_interface $db, \phpbb\template\template $template, \phpbb\admanagement\location\manager $location_manager, $ads_table, $ad_locations_table)
+	public function __construct(\phpbb\request\request $request, \phpbb\db\driver\driver_interface $db, \phpbb\template\template $template, \phpbb\admanagement\ad\manager $manager, \phpbb\admanagement\location\manager $location_manager)
 	{
 		$this->request = $request;
 		$this->db = $db;
```
VSEphpbb: db can be removed from this class

event/main_listener.php
```diff
@@ -77,37 +72,23 @@ public function setup_ads()
 
 		if ($ad_preview)
```
VSEphpbb: For clarity, you can split `setup_ads()` up into to 2 separate methods:
`ad_preview()` and `show_ads()`

These two methods can be assigned to the same event like this:
```php
'core.page_header_after' => array(array('show_ads'), array('preview_ad')),
```

location/type/above_footer.php
```diff
@@ -0,0 +1,54 @@
...
+	* @var \phpbb\user
+	*/
+	protected $user;
+
+	/**
+	* Construct an above_footer template location object
+	*
+	* @param	\phpbb\user	$config	User object
+	*/
+	public function __construct(\phpbb\user $user)
```
VSEphpbb: The constructor should just go in the base class so it doesn't have to be duplicated in every child class

location/type/above_footer.php
```diff
@@ -0,0 +1,54 @@
...
+
+namespace phpbb\admanagement\location\type;
+
+class above_footer extends base
+{
+	/**
+	* User object
+	* @var \phpbb\user
+	*/
+	protected $user;
```
VSEphpbb: Also define this n the base class instead.

location/type/type_interface.php
VSEphpbb: Fair enough then

event/main_listener.php
```diff
@@ -29,56 +29,66 @@ class main_listener implements EventSubscriberInterface
...
+	* @param \phpbb\db\driver\driver_interface		$db					DB driver interface
+	* @param \phpbb\template\template				$template			Template object
+	* @param \phpbb\admanagement\ad\manager			$manager			Advertisement manager object
+	* @param \phpbb\admanagement\location\manager	$location_manager	Template location manager object
 	*/
-	public function __construct(\phpbb\request\request $request, \phpbb\db\driver\driver_interface $db, \phpbb\template\template $template, $ads_table)
+	public function __construct(\phpbb\request\request $request, \phpbb\db\driver\driver_interface $db, \phpbb\template\template $template, \phpbb\admanagement\ad\manager $manager, \phpbb\admanagement\location\manager $location_manager)
 	{
 		$this->request = $request;
 		$this->db = $db;
```
VSEphpbb: the db class is not used in the listener class anymore.

VSEphpbb:
Look at that massive score improvement 👍

ad/manager.php
```diff
@@ -0,0 +1,245 @@
...
+		$this->db->sql_query($sql);
+
+		return $this->db->sql_affectedrows();
+	}
+
+	/**
+	* Get all locations for specified advertisement
+	*
+	* @param	int	$ad_id	Advertisement ID
+	* @return	int	List of template locations for specified ad
```
VSEphpbb: returns an array

controller/admin_controller.php
```diff
@@ -47,21 +50,23 @@ class admin_controller
...
+	* @param \phpbb\request\request					$request			Request object
+	* @param \phpbb\admanagement\ad\manager			$manager			Advertisement manager object
+	* @param \phpbb\admanagement\location\manager	$location_manager	Template location manager object
+	* @param string									$php_ext			PHP extension
+	* @param string									$ext_path			Path to this extension
 	*/
-	public function __construct(\phpbb\db\driver\driver_interface $db, \phpbb\template\template $template, \phpbb\user $user, \phpbb\request\request $request, $ads_table, $php_ext, $ext_path)
+	public function __construct(\phpbb\db\driver\driver_interface $db, \phpbb\template\template $template, \phpbb\user $user, \phpbb\request\request $request, \phpbb\admanagement\ad\manager $manager, \phpbb\admanagement\location\manager $location_manager, $php_ext, $ext_path)
 	{
 		$this->db = $db;
```
VSEphpbb: the db class is no longer needed in this class either.

controller/admin_controller.php
```diff
@@ -154,40 +158,36 @@ public function action_edit()
...
 
 			if (empty($this->errors))
 			{
-				// Insert the ad data to the database
-				$sql = 'UPDATE ' . $this->ads_table . '
-					SET ' . $this->db->sql_build_array('UPDATE', $data) . '
-					WHERE ad_id = ' . (int) $ad_id;
-				$this->db->sql_query($sql);
+				$this->manager->delete_ad_locations($ad_id);
+				$success = $this->manager->update_ad($ad_id, $data);
```
VSEphpbb: Perhaps delete the ad first, and move delete_ad_locations into the if ($success) condition so that way the link locations oly get deleted if the ad was successfully deleted first.

event/main_listener.php
senky: Cool! I always wanted to trigger multiple methods per one event but didn't know it is possible.

Such solution will, however, add duplicate code:

```
public function ad_preview()
{
	$ad_preview = $this->request->variable('ad_preview', 0);
	if ($ad_preview)
	...
}

public function setup_ads()
{
	$ad_preview = $this->request->variable('ad_preview', 0);
	if (!$ad_preview)
	...
}
```

VSEphpbb:
Just ignore the ad preview in show ads method?

senky:
@VSEphpbb see new commit to understand how I intend preview to work and why it is a problem now.

VSEphpbb:
@senky Actually I was going to say to you I think preview's should be done inside the ACP on the page that lists all the ads. 

The ad previews should not be exposed to the "public" or potentially accessible though a simple URL parameter added to the URL (there's a potential they can be abused that way).

ad/manager.php
```diff
@@ -0,0 +1,245 @@
...
+			FROM (
+				SELECT al.location_id, a.ad_code
+				FROM ' . $this->ad_locations_table . ' al
+				LEFT JOIN ' . $this->ads_table . ' a
+					ON (al.ad_id = a.ad_id)
+				WHERE a.ad_enabled = 1
+					AND ' . $this->db->sql_in_set('al.location_id', $ad_locations) . '
+				ORDER BY ' . $this->sql_random() . '
+			) z
+			GROUP BY z.location_id';
```
VSEphpbb: The postgresql failure has this error:
```
ERROR: column "z.ad_code" must appear in the GROUP BY clause or be used in an aggregate function LINE 1: SELECT location_id, ad_code ^ []

SQL

SELECT location_id, ad_code FROM ( SELECT al.location_id, a.ad_code FROM phpbb_ad_locations al LEFT JOIN phpbb_ads a ON (al.ad_id = a.ad_id) WHERE a.ad_enabled = 1 AND al.location_id IN ('above_footer', 'above_header', 'below_footer', 'below_header') ORDER BY RANDOM() ) z GROUP BY z.location_id
```

event/main_listener.php
```diff
@@ -23,62 +23,130 @@ class main_listener implements EventSubscriberInterface
...
-	 * Displays advertisement preview if requested
+	 * Preview advertisement
 	 */
-	public function ad_preview()
+	public function preview_ad()
 	{
 		$ad_preview = $this->request->variable('ad_preview', 0);
+		if ($ad_preview)
+		{
+			$this->user->set_cookie('ad_preview', $ad_preview, time() + 86400);
```
VSEphpbb: Even though as I mentioned I think previewing ads in the user-facing section of the board is not the right way to go (for additional reasons we could discuss on IRC) FYI the cookie thing here is not the right way to handle this. In this sort of situation you should instead only allow someone who is the admin, or has the admin permissions to manage ads to be able to see the previews...so regular users can never see it.

styles/prosilver/template/event/memberlist_view_content_append.html
```diff
@@ -0,0 +1,3 @@
+{% if AD_AFTER_PROFILE %}
+	<div style="text-align: center;">{{ AD_AFTER_PROFILE }}</div>
```
VSEphpbb: using text-align:center does not reliably center content, for example divs won't be centered. We should have a more bullet-proof centering strategy. We should also take responsive-ness into consideration and use a CSS solution that both centers the content and either scales down the ad or hides it gracefully for mobile screens.

ad/manager.php
senky: Postgres always complains about something... I forgot how to solve this. All hints from school's DB course are blown out of my head 😅. Any help with this?

DavidIQ: Wouldn't `ORDER BY z.location_id` provide the same desired result?

senky: You are right! I don't quite understand why it works, but it does. 👍 Thanks!

styles/prosilver/template/event/memberlist_view_content_append.html
senky: I am not fan of sticking into responsiveness. Every ad should handle this itself (AdSense supports responsive ads). Maybe I can give it one more CSS rule to handle simple image banners to scale down. But there is no general rule when to scale and when to hide.

controller/admin_controller.php
```diff
@@ -114,14 +114,20 @@ public function get_page_title()
 	*/
 	public function action_add()
 	{
+		$preview = $this->request->is_set_post('preview');
+
```
VSEphpbb: can you add a `$submit = ...` here as well. It's more readable when you have `if ($preview || $submit)`

adm/style/manage_ads.html
```diff
@@ -8,6 +8,12 @@
 
 	<h1>{% if S_ADD_AD %}{{ lang('ACP_ADS_ADD') }}{% else %}{{ lang('ACP_ADS_EDIT') }}{% endif %}</h1>
 
+	{% if PREVIEW %}
+		<div style="text-align: center;">
```
VSEphpbb: This should use the ad-center class as well

senky:
!unset WIP

senky:
!set missing tests

VSEphpbb:
So many ads. I think we could use a few pixels of padding above/below the ads?
screen shot 2017-06-02 at 8 03 19 am

VSEphpbb:
Well they better not do that then. Some ad codes are scripts (like google adsense) so scripts have to be allowed.

ad/manager.php
```diff
@@ -0,0 +1,245 @@
...
+
+		return $this->db->sql_nextid();
+	}
+
+	/**
+	* Update advertisement
+	*
+	* @param	int		$ad_id	Advertisement ID
+	* @param	array	$data	List of data to update in the database
+	* @return	int		Number of affected rows. Can be used to determine if any ad has been udated.
```
VSEphpbb: `udated` spelling

controller/admin_controller.php
```diff
@@ -114,24 +114,29 @@ public function get_page_title()
...
+			}
+			else if (empty($this->errors))
 			{
-				// Insert the ad data to the database
-				$sql = 'INSERT INTO ' . $this->ads_table . ' ' . $this->db->sql_build_array('INSERT', $data);
-				$this->db->sql_query($sql);
+				$ad_id = $this->manager->insert_ad($data);
+				$this->manager->insert_ad_locations($ad_id, $data['ad_locations']);
 
 				$this->success('ACP_AD_ADD_SUCCESS');
```
VSEphpbb: This addition should be logged

controller/admin_controller.php
```diff
@@ -140,6 +145,7 @@ public function action_add()
 			'S_ADD_AD'	=> true,
 			'U_BACK'	=> $this->u_action,
 		));
+		$this->assign_locations();
```
VSEphpbb: Does this overwrite the assign_locations called in line 139?

controller/admin_controller.php
```diff
@@ -150,44 +156,46 @@ public function action_add()
...
+				$success = $this->manager->update_ad($ad_id, $data);
 
-				if ($this->db->sql_affectedrows())
+				if ($success)
 				{
+					// Only insert new ad locations to DB when ad exists
+					$this->manager->delete_ad_locations($ad_id);
+					$this->manager->insert_ad_locations($ad_id, $data['ad_locations']);
+
 					$this->success('ACP_AD_EDIT_SUCCESS');
```
VSEphpbb: Log this action too

controller/admin_controller.php
```diff
@@ -231,12 +240,11 @@ public function action_delete()
...
 			{
-				$sql = 'DELETE FROM ' . $this->ads_table . '
-					WHERE ad_id = ' . (int) $ad_id;
-				$this->db->sql_query($sql);
+				$this->manager->delete_ad_locations($ad_id);
+				$success = $this->manager->delete_ad($ad_id);
 
 				// Only notify user on error or if not ajax
-				if (!$this->db->sql_affectedrows())
+				if (!$success)
```
VSEphpbb: Lets also log successful deleted ads

language/en/common.php
```diff
@@ -19,15 +19,5 @@
...
-	'DEMO_GOODBYE'		=> 'Goodbye %s!',
-
-	'ACP_DEMO'					=> 'Settings',
-	'ACP_DEMO_GOODBYE'			=> 'Should say goodbye?',
-	'ACP_DEMO_SETTING_SAVED'	=> 'Settings have been saved successfully!',
-
-	'ACME_DEMO_NOTIFICATION'	=> 'Acme demo notification',
-
-	'VIEWING_ACME_DEMO'			=> 'Viewing Acme Demo',
+	'PREVIEWING_AD'	=> 'You are currently previewing the advertisement. You can browse the board and see how it looks on various locations. To stop previewing the ad, <a href="%s">click here</a>.',
```
VSEphpbb: is this still needed?

migrations/v10x/m3_template_locations_schema.php
```diff
@@ -0,0 +1,70 @@
...
+	*/
+	static public function depends_on()
+	{
+		return array('\phpbb\admanagement\migrations\v10x\m2_acp_module');
+	}
+
+	/**
+	* Add the ad_locations table schema to the database:
+	*	ad_locations:
+	*		ad_location_id
```
VSEphpbb: ad_location_id?? typo??

senky:
!unset missing tests

rxu:
@senky Nice, thanks.

senky:
@VSEphpbb @Derky I am thinking that since we will submit the extension to CDB after this PR, it should already contain DB indices to speed up ad fetching query that is triggered on every board hit...

styles/all/theme/admanagement.css
```diff
@@ -1,4 +1,6 @@
 .ad-center {
 	display: flex;
 	justify-content: center;
+
```
VSEphpbb: This blank line should be removed

VSEphpbb:
@senky that is fine...don't you already have all the indexes you need?

Also, before you can submit this to CDB you need to update the minimum phpBB version to 3.1.6.

1 reason is because the way we use INCLUDECSS, that syntax was not available until 3.1.2. And then some of the template events you are using should be checked in phpbb/docs/events.md. You should double check but it looks like the newest event we are using was added in 3.1.6.

If you want to make sure people can only install the ext on >= 3.1.6, then add an ext.php file with an is_enabledable test (like may of the other official extensions do).

ext.php
```diff
@@ -0,0 +1,22 @@
...
+class ext extends \phpbb\extension\base
+{
+	/**
+	 * {@inheritdoc}
+	 */
+	public function is_enableable()
+	{
+		return phpbb_version_compare(PHPBB_VERSION, '3.1.6', '>=');
+	}
+}
```
VSEphpbb: Missing new line at EOF

tests/event/main_listener_base.php
```diff
@@ -47,11 +44,19 @@ public function setUp()
...
+		$this->ads_table = 'phpbb_ads';
+		$this->ad_locations_table = 'phpbb_ad_locations';
+
 		// Load/Mock classes required by the listener class
-		$this->request = $this->getMock('\phpbb\request\request');
-		$this->db = $this->new_dbal();
 		$this->template = $this->getMock('\phpbb\template\template');
-		$this->ads_table = 'phpbb_ads';
+		$this->manager = new \phpbb\admanagement\ad\manager($this->new_dbal(), $this->ads_table, $this->ad_locations_table);
+		$this->location_manager = new \phpbb\admanagement\location\manager(array(
```
VSEphpbb: You could improve test coverage if you load all the built-in location types. Here's a way to achieve that:
```php
// Location types
$locations = array('above_footer', 'above_header', 'after_first_post', 
                   'after_not_first_post', 'after_posts', 'after_profile',
                   'before_posts', 'before_profile', 'below_footer', 'below_header');
$location_types = array();
foreach ($locations as $type)
{
	$class = "\\phpbb\\admanagement\\location\\type\\$type";
	$location_types['phpbb.admanagement.location.type.' . $type] = new $class($user);
}

// Location manager
$this->location_manager = new \phpbb\admanagement\location\manager($location_types_array);
```

PR #5, opened 2017-06-06
End Date

senky:
Admin can set date after which advertisement will be automatically disabled.


senky:
!set WIP

VSEphpbb:
I assume you already have a plan for setting the expiry date, but if you want to see some other ideas in action, take a look at how we implemented setting an expiry date in the Board Announcements extension (we implemented an external JS script for a date-picker) https://github.com/phpbb-extensions/boardannouncements/pull/111

senky:
@VSEphpbb That's what I needed! I was just wrapping my head around it.

VSEphpbb:
@senky  If you look at board announcements, it handles the date in a format of YYYY-MM-DD or something, but in PHP we transform it to a timestamp for storage in the db. So it should be pretty straightforward how to implement it.

senky:
!unset WIP

senky:
!set missing tests

adm/style/event/acp_overall_footer_after.html
```diff
@@ -20,3 +23,13 @@
...
+
+jQuery(function($) {
+	$('#ad_end_date').datetimepicker({
+		format:'{{ PICKER_DATE_FORMAT|e('js') }}',
+		validateOnBlur: false,
+		minDate: 0
+	});
+	$.datetimepicker.setLocale('{S_USER_LANG}');
+});
+</script>
```
VSEphpbb: All of this should only be loaded if we're in our ACP panel only (currently it's being loaded globally in ACP)

adm/style/event/acp_overall_header_head_append.html
```diff
@@ -1 +1,2 @@
 {% INCLUDECSS '@phpbb_admanagement/admanagement.css' %}
+{% INCLUDECSS '@phpbb_admanagement/datetimepicker/jquery.datetimepicker.min.css' %}
```
VSEphpbb: All these scripts should only be loaded in our ACP too (currently being loaded everywhere)

adm/style/event/acp_overall_footer_after.html
```diff
@@ -20,3 +23,13 @@
...
 	$this.attr('href', newHref);
 });
+
+jQuery(function($) {
+	$('#ad_end_date').datetimepicker({
+		format:'{{ PICKER_DATE_FORMAT|e('js') }}',
+		validateOnBlur: false,
+		minDate: 0
+	});
+	$.datetimepicker.setLocale('{S_USER_LANG}');
```
VSEphpbb: This should be a twig variable `{{ S_USER_LANG }}`

language/en/acp.php
```diff
@@ -21,12 +21,14 @@
 $lang = array_merge($lang, array(
 	'AD_NAME'				=> 'Name',
 	'AD_NAME_EXPLAIN'		=> 'This is only used for your recognition of this advertisement.',
+	'AD_ENABLED'			=> 'Enabled',
+	'AD_ENABLED_EXPLAIN'	=> 'If disabled, this advertisement will not be displayed.',
+	'AD_END_DATE'			=> 'End Date',
+	'AD_END_DATE_EXPLAIN'	=> 'Set the date the advertisement will expire and become disabled. Leave this field blank if you do not want the advertisement to expire. Please, use <samp>YYYY-MM-DD HH:MM</samp> format.',
```
VSEphpbb: I think for this extension we don't need time, just a date. There are settings for the date picker to disable choosing a time of day.

Set `timepicker` to `false` and update the format accordingly to just Y-m-d

adm/style/event/acp_overall_footer_after.html
VSEphpbb: Actually this can still be an external JS script file as it was previously. Just move the INCLUDES for the datetimepicker into our manage_ads.html page

migrations/v10x/m5_end_date.php
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * @copyright (c) 2017 phpBB Limited <https://www.phpbb.com>
+ * @license GNU General Public License, version 2 (GPL-2.0)
+ *
+ */
+
+namespace phpbb\admanagement\migrations\v10x;
+
+/**
+* Migration stage 1: Initial schema
```
VSEphpbb: Update this comment please

adm/style/event/acp_overall_footer_after.html
```diff
@@ -1,3 +1,6 @@
+{% INCLUDEJS '@phpbb_admanagement/datetimepicker/jquery.datetimepicker.full.min.js' %}
+
+<script>
```
VSEphpbb: A `;` is required after the use strict statement below

adm/style/event/acp_overall_footer_after.html
VSEphpbb: Oh sorry I see why you had to put this into HTML. Disregard my previous comment ;)

adm/style/event/acp_overall_footer_after.html
```diff
@@ -1,3 +1,6 @@
+{% INCLUDEJS '@phpbb_admanagement/datetimepicker/jquery.datetimepicker.full.min.js' %}
```
VSEphpbb: Also to prevent multiple loads of 3rd party JS/CSS scripts we recommend doing something like this:

```twig
{% if not definition.INCLUDED_DATETIMEPICKERJS %}
    {% INCLUDEJS '@phpbb_boardannouncements/datetimepicker/jquery.datetimepicker.full.min.js' %}
    {% DEFINE INCLUDED_DATETIMEPICKERJS = true %}
{% endif %}
```

senky:
@VSEphpbb End Date explanation says, that advertisement will become disabled. I am thinking of cron task that will perform this (say once a day) to ease the main query.
A drawback is inconsistency displayed to the admin in ACP. Some ads that are expired will be disabled and some (temporarily) enabled.

So maybe the best solution is to add a new column to ACP list to signalize (say with red font color) expired ads?

senky:
!unset missing tests

adm/style/event/acp_overall_footer_after.html
```diff
@@ -1,10 +1,17 @@
+{% if not definition.INCLUDED_DATETIMEPICKERJS and (S_ADD_AD or S_EDIT_AD) %}
```
VSEphpbb: I would just put this entire page inside an S_ADS switch that means we are in our ACP page... I wouldn't bother distibguishing between the add or edit. We can just load it all whether we're adding or editing

adm/style/event/acp_overall_header_head_append.html
```diff
@@ -1 +1,4 @@
-{% INCLUDECSS '@phpbb_admanagement/admanagement.css' %}
+{% if S_ADS or S_ADD_AD or S_EDIT_AD %}
```
VSEphpbb: Also here, just a single S_ADS switch will be enough

VSEphpbb:
@senky Cron is an option but it's not always immediate. Instead of CRON, what I think we should do is add a method when we load our ACP page that will immediately disable any expired ads, so they are displayed to the admin correctly. I think the query itself is OK the way you have it.

VSEphpbb:
@Senky also yes, a column indicating an ad has expired would be nice and informative to the users as well.

VSEphpbb:
Maybe say "Expired on 1-1-2018" for example

adm/style/event/acp_overall_footer_after.html
```diff
@@ -1,42 +1,41 @@
...
-		timepicker: false,
+		$this.attr('href', newHref);
+	});
+
+	jQuery(function($) {
+		$('#ad_end_date').datetimepicker({
+			format:'{{ PICKER_DATE_FORMAT|e('js') }}',
+			validateOnBlur: false,
+			minDate: 0,
+			timepicker: false,
```
VSEphpbb: the comma is not needed on the last item

controller/admin_controller.php
```diff
@@ -81,6 +81,12 @@ public function main()
 	{
 		$this->user->add_lang_ext('phpbb/admanagement', 'acp');
 
+		$this->manager->disable_expired_ads();
+
+		$this->template->assign_vars(array(
+			'S_PHPBB_ADMANAGEMENT'	=> true,
```
VSEphpbb: when assigning a single var you can use 
`template->assign_var('S_PHPBB_ADMANAGEMENT', true)`
There is another spot where this can be updated in this file as well

language/en/acp.php
```diff
@@ -35,10 +35,11 @@
 		0 => 'Click to enable',
 		1 => 'Click to disable',
 	),
-	'ACP_ADS_EMPTY'		=> 'No advertisement, yet. Add one with the button below.',
-	'ACP_ADS_ADD'		=> 'Add new advertisement',
-	'ACP_ADS_EDIT'		=> 'Edit advertisement',
-	'AD_PREVIEW'		=> 'Preview this advertisement',
+	'AD_EXPIRED_EXPLAIN'	=> 'This advertisement is past end date. We have disabled it automatically.',
```
VSEphpbb: I would prefer saying:
"This advertisement has expired and has been disabled."

adm/style/manage_ads.html
```diff
@@ -84,6 +85,7 @@
 			{% for ad in loops.ads %}
 				<tr>
 					<td><strong>{{ ad.NAME }}</strong></td>
+					<td>{% if ad.END_DATE_EXPIRED %}<strong title="{{ lang('AD_EXPIRED_EXPLAIN') }}" style="color: red;">{% endif %}{{ ad.END_DATE }}{% if ad.END_DATE_EXPIRED %}</strong>{% endif %}</td>
```
VSEphpbb: You can use class="error" instead of style="red"

ad/manager.php
```diff
@@ -148,6 +148,19 @@ public function delete_ad($ad_id)
...
 	/**
+	* Disable expired advertisements
+	*
+	* @return	void
+	*/
+	public function disable_expired_ads()
+	{
+		$sql = 'UPDATE ' . $this->ads_table . '
+			SET ad_enabled = 0
+			WHERE ad_end_date < ' . time();
```
VSEphpbb: This is disabling ads with no expiration set (because it is 0)

controller/admin_controller.php
```diff
@@ -301,17 +307,18 @@ public function list_ads()
...
 
 			$this->template->assign_block_vars('ads', array(
-				'NAME'		=> $row['ad_name'],
-				'S_ENABLED'	=> $ad_enabled,
-				'U_ENABLE'	=> $this->u_action . '&amp;action=' . ($ad_enabled ? 'disable' : 'enable') . '&amp;id=' . $row['ad_id'],
-				'U_EDIT'	=> $this->u_action . '&amp;action=edit&amp;id=' . $row['ad_id'],
-				'U_DELETE'	=> $this->u_action . '&amp;action=delete&amp;id=' . $row['ad_id'],
+				'NAME'				=> $row['ad_name'],
+				'END_DATE'			=> $row['ad_end_date'] ? $this->user->format_date($row['ad_end_date'], self::DATE_FORMAT) : '',
+				'END_DATE_EXPIRED'	=> (int) $row['ad_end_date'] < time(),
```
VSEphpbb: This evaluates to true for ads with no set end date, which should not be the case.

VSEphpbb: also END_DATE_EXPIRED is being used as a boolean switch so it should start with S_

adm/style/manage_ads.html
```diff
@@ -85,8 +85,16 @@
 			{% for ad in loops.ads %}
 				<tr>
 					<td><strong>{{ ad.NAME }}</strong></td>
-					<td>{% if ad.END_DATE_EXPIRED %}<strong title="{{ lang('AD_EXPIRED_EXPLAIN') }}" style="color: red;">{% endif %}{{ ad.END_DATE }}{% if ad.END_DATE_EXPIRED %}</strong>{% endif %}</td>
-					<td><a href="{{ ad.U_ENABLE }}" title="{{ lang('AD_ENABLE_TITLE', ad.S_ENABLED) }}" data-ajax="toggle_enable">{% if ad.S_ENABLED %}{{ lang('ENABLED') }}{% else %}{{ lang('DISABLED') }}{% endif %}</a></td>
+					<td>
+						{% if ad.END_DATE_EXPIRED %}<strong title="{{ lang('AD_EXPIRED_EXPLAIN') }}" class="error">{% endif %}
```
VSEphpbb: I don't think this needs to be strong as the red color is pretty significant a stand-out effect already.

controller/admin_controller.php
```diff
@@ -303,11 +301,18 @@ public function list_ads()
...
+				$ad_enabled = 0;
+				$this->manager->update_ad($row['ad_id'], array('ad_enabled' => 0));
+			}
 
 			$this->template->assign_block_vars('ads', array(
 				'NAME'				=> $row['ad_name'],
-				'END_DATE'			=> $row['ad_end_date'] ? $this->user->format_date($row['ad_end_date'], self::DATE_FORMAT) : '',
-				'END_DATE_EXPIRED'	=> (int) $row['ad_end_date'] < time(),
+				'END_DATE'			=> $ad_end_date ? $this->user->format_date($ad_end_date, self::DATE_FORMAT) : '',
+				'END_DATE_EXPIRED'	=> $ad_end_date_condition,
```
VSEphpbb: S_END_DATE_EXPIRED

adm/style/event/acp_overall_footer_after.html
```diff
@@ -0,0 +1,41 @@
+{% if not definition.INCLUDED_DATETIMEPICKERJS and S_PHPBB_ADMANAGEMENT %}
```
VSEphpbb: I think this should go inside the S_PHPBB_ADMANAGEMENT below so we're not checking this condition twice ;)

controller/admin_controller.php
```diff
@@ -296,20 +301,27 @@ public function list_ads()
 		foreach ($this->manager->get_all_ads() as $row)
 		{
 			$ad_enabled = (int) $row['ad_enabled'];
+			$ad_end_date = (int) $row['ad_end_date'];
+			$ad_expired = $ad_end_date > 0 && $ad_end_date < time();
+			if ($ad_expired)
```
VSEphpbb: Actually this condition should be `if ($ad_expired && !$ad_disabled)` so we don't keep disabling already disabled expired ads.

VSEphpbb: I mean !$ad_enabled but you knew that

PR #12, opened 2017-06-09
#10: Merge get_form_data() and validate()

### LINKED ISSUE #10: Merge get_form_data() and validate()
This will enable to test end date input properly.

senky:
Resolves #10

### DIFF
--- a/controller/admin_controller.php
+++ b/controller/admin_controller.php
@@ -128,9 +128,7 @@ class admin_controller
 		add_form_key('phpbb/admanagement/add');
 		if ($preview || $submit)
 		{
-			$data = $this->get_form_data();
-
-			$this->validate($data, 'phpbb/admanagement/add');
+			$data = $this->get_form_data('phpbb/admanagement/add');
 
 			if ($preview)
 			{
@@ -176,9 +174,7 @@ class admin_controller
 		add_form_key('phpbb/admanagement/edit/' . $ad_id);
 		if ($preview || $submit)
 		{
-			$data = $this->get_form_data();
-
-			$this->validate($data, 'phpbb/admanagement/edit/' . $ad_id);
+			$data = $this->get_form_data('phpbb/admanagement/edit/' . $ad_id);
 
 			if ($preview)
 			{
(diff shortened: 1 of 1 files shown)


controller/admin_controller.php
```diff
@@ -398,10 +387,27 @@ protected function validate($data, $form_name)
...
+
+		// Validate ad end date pattern
+		if (!preg_match('#(^\d{4}\-\d{2}\-\d{2}$)?#', $data['ad_end_date']))
+		{
+			$this->errors[] = $this->user->lang('AD_END_DATE_INVALID');
+
+			// Return immediately to retain end date text.
+			// Admin might just accidentally remove one number, so we
+			// don't want to remove it all.
+			return $data;
```
VSEphpbb: My only problem with this is we are forced to make sure we don't add any additional validations after this. I would rather instead of returning $data early here, instead just use an `else` condition to further prepare ad_end_date...so we only ever return $data once, at the end of this method.

VSEphpbb: Maybe this can all be simplified to something like:
```php
if (preg_match('#(^\d{4}\-\d{2}\-\d{2}$)?#', $data['ad_end_date']))
{
	$data['ad_end_date'] = (int) $this->user->get_timestamp_from_format(self::DATE_FORMAT, $data['ad_end_date']);

	if ($data['ad_end_date'] != 0 && $data['ad_end_date'] < time())
	{
		$this->errors[] = $this->user->lang('AD_END_DATE_INVALID');
	}
}
else
{
	$this->errors[] = $this->user->lang('AD_END_DATE_INVALID');
}
```

VSEphpbb: I updated my example. It's maybe not the best but this does work too. Notice I changed the reggae pattern a little bit.
```php
if (preg_match('#^\d{4}\-\d{2}\-\d{2}$#', $data['ad_end_date']))
{
	$data['ad_end_date'] = (int) $this->user->get_timestamp_from_format(self::DATE_FORMAT, $data['ad_end_date']);

	if ($data['ad_end_date'] < time())
	{
		$this->errors[] = $this->user->lang('AD_END_DATE_INVALID');
	}
}
else if ($data['ad_end_date'] !== '')
{
	$this->errors[] = $this->user->lang('AD_END_DATE_INVALID');
}
else
{
	$data['ad_end_date'] = 0;
}
```

senky:
@VSEphpbb to be honest, I don't like how this validation works at all. It can be string or int, data source is DB some times, request variable other times. It makes validation too complicated.

Why we can't use DATE column type in DB? Conversion from string to unix timestamp is the problem here...

VSEphpbb:
@senky I don't see it as a problem. The user input is always a string for this field. So first we validate if the string is what we expect, and if it is, then convert it to the timestamp, check that and so on. I'm just looking for a clear and concide few lines sthat do that validation in a clean way.

PR #14, opened 2017-06-10
Try testing the ACP module

VSEphpbb:


ad/manager.php
```diff
@@ -39,7 +39,7 @@ public function __construct(\phpbb\db\driver\driver_interface $db, $ads_table, $
 	* Get specific ad
 	*
 	* @param	int		$ad_id	Advertisement ID
-	* @return	array	Advertisement data
+	* @return	mixed	Array with advertisement data, false if ad doesn't exist
```
senky: This probably doesn't belong here...

senky:
You really want to achieve 99% code coverage, don't you? 😄

ad/manager.php
VSEphpbb: Well, we are returning what comes from fetchrow, which is either gonna be an array or false.

VSEphpbb: Gotta fix it somewhere 😄

PR #16, opened 2017-06-10
Hide advertisement for specific group

senky:
Admin can specify groups advertisement will not be displayed to.

### DIFF
--- a/ad/manager.php
+++ b/ad/manager.php
@@ -233,7 +233,8 @@ class manager
 	public function load_groups()
 	{
 		$sql = 'SELECT group_id, group_name
-			FROM ' . GROUPS_TABLE;
+			FROM ' . GROUPS_TABLE . "
+			ORDER BY group_name ASC";
 		$result = $this->db->sql_query($sql);
 		$groups = $this->db->sql_fetchrowset($result);
 		$this->db->sql_freeresult($result);
--- a/ad/manager.php
+++ b/ad/manager.php
@@ -204,6 +204,44 @@ class manager
 	}
 
 	/**
+	* Load memberships of the user
+	*
+	* @param	int		$user_id	User ID to load memberships
+	* @return	array	List of group IDs user is member of
+	*/
+	public function load_memberships($user_id)
+	{
+		$memberships = array();
+		$sql = 'SELECT group_id
+			FROM ' . USER_GROUP_TABLE . '
+			WHERE user_id = ' . (int) $user_id . '
+			AND user_pending = 0';
+		$result = $this->db->sql_query($sql, 86400);
+		while ($row = $this->db->sql_fetchrow($result))
+		{
+			$memberships[] = $row['group_id'];
+		}
+		$this->db->sql_freeresult($result);
+		return $memberships;
+	}
+
+	/**
+	* Load all board groups
+	*
+	* @return	array	List of groups
+	*/
+	public function load_groups()
+	{
+		$sql = 'SELECT group_id, group_name
+			FROM ' . GROUPS_TABLE;
+		$result = $this->db->sql_query($sql);
+		$groups = $this->db->sql_fetchrowset($result);
+		$this->db->sql_freeresult($result);
+
+		return $groups;
+	}
+
+	/**
 	* Make sure only necessary data make their way to SQL query
 	*
 	* @param	array	$data	List of data to query the database
--- a/ad/manager.php
+++ b/ad/manager.php
@@ -216,7 +216,7 @@ class manager
 			FROM ' . USER_GROUP_TABLE . '
 			WHERE user_id = ' . (int) $user_id . '
 			AND user_pending = 0';
-		$result = $this->db->sql_query($sql, 86400);
+		$result = $this->db->sql_query($sql, 3600);
 		while ($row = $this->db->sql_fetchrow($result))
 		{
 			$memberships[] = $row['group_id'];
--- a/ad/manager.php
+++ b/ad/manager.php
@@ -232,7 +232,7 @@ class manager
 	*/
 	public function load_groups()
 	{
-		$sql = 'SELECT group_id, group_name
+		$sql = 'SELECT group_id, group_name, group_type
 			FROM ' . GROUPS_TABLE . "
 			ORDER BY group_name ASC";
 		$result = $this->db->sql_query($sql);
--- a/controller/admin_controller.php
+++ b/controller/admin_controller.php
@@ -140,7 +140,7 @@ class admin_controller
 		$groups = $this->manager->load_groups();
 		foreach ($groups as $group)
 		{
-			$group_name = (!empty($this->user->lang['G_' . $group['group_name']])) ? $this->user->lang('G_' . $group['group_name']) : $group['group_name'];
+			$group_name = ($group['group_type'] == GROUP_SPECIAL) ? $this->user->lang('G_' . $group['group_name']) : $group['group_name'];
 
 			$this->template->assign_block_vars('groups', array(
 				'ID'			=> $group['group_id'],
--- a/controller/admin_controller.php
+++ b/controller/admin_controller.php
@@ -118,13 +118,23 @@ class admin_controller
 			// Validate form key
 			if (!check_form_key('phpbb/admanagement/settings'))
 			{
-				$errors[] = $this->user->lang('FORM_INVALID');
+				$this->errors[] = $this->user->lang('FORM_INVALID');
 			}
 
-			$this->config_text->set('phpbb_admanagement_hide_groups', json_encode($this->request->variable('hide_groups', array(0))));
-			$this->cache->destroy('sql', USER_GROUP_TABLE);
+			if (empty($this->errors))
+			{
+				$this->config_text->set('phpbb_admanagement_hide_groups', json_encode($this->request->variable('hide_groups', array(0))));
+				$this->cache->destroy('sql', USER_GROUP_TABLE);
 
-			$this->success('ACP_AD_SETTINGS_SAVED');
+				$this->success('ACP_AD_SETTINGS_SAVED');
+			}
+			else
+			{
+				$this->template->assign_vars(array(
+					'S_ERROR'		=> (bool) count($this->errors),
+					'ERROR_MSG'		=> count($this->errors) ? implode('<br />', $this->errors) : '',
+				));
+			}
 		}
 
 		$hide_groups = json_decode($this->config_text->get('phpbb_admanagement_hide_groups'), true);
--- a/controller/admin_controller.php
+++ b/controller/admin_controller.php
@@ -127,6 +127,8 @@ class admin_controller
 			}
 
 			$this->config->set('phpbb_admanagement_hide_groups', implode(',', $this->request->variable('hide_groups', array(0))));
+
+			$this->success('ACP_AD_SETTINGS_SAVED');
 		}
 
 		$hide_groups = explode(',', $this->config['phpbb_admanagement_hide_groups']);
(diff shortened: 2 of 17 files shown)


VSEphpbb:
For the failing test you need to update the acp module test. You need to give the last test a data array with the available modes (manage and settings)
```php
return array(
	array('manage'),
	array('settings'),
);
```
and update the test itself like 
```php
$admin_controller
	->expects($this->once())
	->method("mode_$mode");

$p_master = new p_master();
$p_master->load('acp', '\phpbb\admanagement\acp\main_module', $mode);
```

You also need to update the test with the new acp_info data

event/main_listener.php
```diff
@@ -43,28 +55,51 @@ static public function getSubscribedEvents()
...
+		if (!function_exists('group_memberships'))
+		{
+			include($this->root_path . 'includes/functions_user.' . $this->php_ext);
+		}
+		$user_groups = group_memberships(false, $this->user->data['user_id']);
+		$user_groups = array_map(function ($group)
+		{
+			return $group['group_id'];
+		}, $user_groups);
+		$hide_groups = explode(',', $this->config['phpbb_admanagement_hide_groups']);
```
VSEphpbb: This all seems like a really heavy load to run on every single page load. Just on my simple test board it sucks up a whole megabyte of extra ram.

Can we just simply check if the `$user->data['group_id']` is in the list of hide groups? This would of course only be based on the user's main/default group but I believe that is enough.

If not, then I would suggest we have our own group membership query that is simple:
```
		$sql = 'SELECT group_id
			FROM ' . USER_GROUP_TABLE . '
			WHERE user_id = ' . (int) $this->user->data['user_id'] . '
			AND user_pending = 0';
		while ($row = $this->db->sql_fetchrow($result))
		{
			$memberships[] = $row['group_id'];
		}
```

Also I'd like to see as much of this code/logic actually take place in some dedicated class (I guess in our ad class) rather than inside the listener.

controller/admin_controller.php
```diff
@@ -95,6 +108,46 @@ public function main()
...
+		add_form_key('phpbb/admanagement/settings');
+		if ($this->request->is_set_post('submit'))
+		{
+			// Validate form key
+			if (!check_form_key('phpbb/admanagement/settings'))
+			{
+				$errors[] = $this->user->lang('FORM_INVALID');
+			}
+
+			$this->config->set('phpbb_admanagement_hide_groups', implode(',', $this->request->variable('hide_groups', array(0))));
```
VSEphpbb: When storing this data, it is better to both:
1. json_encode the array for storage, and decode for usage instead of a comma delineated string
2. use the config_text table. Because there will be users who will have so many groups they will exceed the 255 char limit of the config table.

controller/admin_controller.php
```diff
@@ -95,6 +108,46 @@ public function main()
...
+
+			$this->success('ACP_AD_SETTINGS_SAVED');
+		}
+
+		$hide_groups = explode(',', $this->config['phpbb_admanagement_hide_groups']);
+		$sql = 'SELECT group_id, group_name
+			FROM ' . $this->groups_table;
+		$result = $this->db->sql_query($sql);
+		while ($row = $this->db->sql_fetchrow($result))
```
VSEphpbb: I'd prefer all DB stuff hapen outside of this class, I guess in the ads class. Unless you see a need for a new class to handle/manage this stuff (`groups`??).

event/main_listener.php
VSEphpbb: I just tested the query I sampled above and it's significantly less strenous than the group_memberships one. We should also probably cache the query so it's even faster as the user browses. But again we should only use this query if we think that the user's default group is not enough, and we need to care about every group the user is a member of.

senky:
!set missing tests

ad/manager.php
```diff
@@ -204,6 +204,44 @@ public function delete_ad_locations($ad_id)
...
+	* @return	array	List of group IDs user is member of
+	*/
+	public function load_memberships($user_id)
+	{
+		$memberships = array();
+		$sql = 'SELECT group_id
+			FROM ' . USER_GROUP_TABLE . '
+			WHERE user_id = ' . (int) $user_id . '
+			AND user_pending = 0';
+		$result = $this->db->sql_query($sql, 86400);
```
VSEphpbb: I think 24 hours is too long. I think about big boards that may have hundreds or thousands of daily visitors. Their cache will be flooded. I don't know how big a problem that is, but I think this should probably be 1 hour instead of cache time.

ad/manager.php
```diff
@@ -204,6 +204,44 @@ public function delete_ad_locations($ad_id)
...
+	/**
+	* Load all board groups
+	*
+	* @return	array	List of groups
+	*/
+	public function load_groups()
+	{
+		$sql = 'SELECT group_id, group_name
+			FROM ' . GROUPS_TABLE;
+		$result = $this->db->sql_query($sql);
```
VSEphpbb: Let's use this query:
```sql
$sql = 'SELECT group_id, group_name
	FROM ' . GROUPS_TABLE . "
	WHERE group_name <> 'BOTS'
	ORDER BY group_name ASC";
```
It will sort them better and also eliminate useless groups like BOTS

controller/admin_controller.php
```diff
@@ -95,6 +103,45 @@ public function main()
...
 
 	/**
+	* Process user request for settings mode
+	*
+	* @return void
+	*/
+	public function mode_settings()
+	{
+		$this->setup();
+		$this->user->add_lang('common');
```
VSEphpbb: what is this for?

controller/admin_controller.php
```diff
@@ -95,6 +103,45 @@ public function main()
...
+		if ($this->request->is_set_post('submit'))
+		{
+			// Validate form key
+			if (!check_form_key('phpbb/admanagement/settings'))
+			{
+				$errors[] = $this->user->lang('FORM_INVALID');
+			}
+
+			$this->config_text->set('phpbb_admanagement_hide_groups', json_encode($this->request->variable('hide_groups', array(0))));
+			$this->cache->destroy('sql', USER_GROUP_TABLE);
```
VSEphpbb: What is the reason for destroying the user group table query cache?

migrations/v10x/m6_hide_for_group.php
```diff
@@ -0,0 +1,51 @@
...
+namespace phpbb\admanagement\migrations\v10x;
+
+class m6_hide_for_group extends \phpbb\db\migration\migration
+{
+	/**
+	* {@inheritDoc}
+	*/
+	public function effectively_installed()
+	{
+		return isset($this->config['phpbb_admanagement_hide_groups']);
```
VSEphpbb: this needs to be corrected. you can maybe do it by checking:
```php
return $config_text->get(phpbb_admanagement_hide_groups) !== null;
```

controller/admin_controller.php
```diff
@@ -95,6 +103,45 @@ public function main()
...
+		$this->setup();
+		$this->user->add_lang('common');
+
+		add_form_key('phpbb/admanagement/settings');
+		if ($this->request->is_set_post('submit'))
+		{
+			// Validate form key
+			if (!check_form_key('phpbb/admanagement/settings'))
+			{
+				$errors[] = $this->user->lang('FORM_INVALID');
```
VSEphpbb: `$this->errors[]`

controller/admin_controller.php
```diff
@@ -95,6 +103,45 @@ public function main()
...
+			$this->cache->destroy('sql', USER_GROUP_TABLE);
+
+			$this->success('ACP_AD_SETTINGS_SAVED');
+		}
+
+		$hide_groups = json_decode($this->config_text->get('phpbb_admanagement_hide_groups'), true);
+		$groups = $this->manager->load_groups();
+		foreach ($groups as $group)
+		{
+			$group_name = (!empty($this->user->lang['G_' . $group['group_name']])) ? $this->user->lang('G_' . $group['group_name']) : $group['group_name'];
```
VSEphpbb: Use `lang()` instead of `lang[]`

ad/manager.php
senky: I believe bots should be included. Admin might want to hide ads for bots to make views/clicks more precise.

controller/admin_controller.php
senky: Pre-defined group names.

controller/admin_controller.php
senky: To immediately reflect ad management settings on frontend. Imagine someone sets ads to hide for admins group, visits index and sees the ads. Must confuse him very much.

controller/admin_controller.php
senky: I really would like to, but I can't test existence of key using `lang()`.

ad/manager.php
VSEphpbb: OK

controller/admin_controller.php
VSEphpbb: You could instead do this test then:
```
$group_name = ($group['group_type'] == GROUP_SPECIAL) ?  ? $this->user->lang('G_' . $group['group_name']) : $group['group_name'];
```
Just make sure to ad `group_type` to the query

controller/admin_controller.php
VSEphpbb: This is not needed. Common is always already loaded.

controller/admin_controller.php
VSEphpbb: But our ad management settings only take place in the config_text table...nothing changes regarding data from the user group table and what groups users belong to. Without this line the ad group view settings are still immediately being reflected (for me).

controller/admin_controller.php
```diff
@@ -118,13 +118,23 @@ public function mode_settings()
...
-			$this->cache->destroy('sql', USER_GROUP_TABLE);
+			if (empty($this->errors))
+			{
+				$this->config_text->set('phpbb_admanagement_hide_groups', json_encode($this->request->variable('hide_groups', array(0))));
+				$this->cache->destroy('sql', USER_GROUP_TABLE);
 
-			$this->success('ACP_AD_SETTINGS_SAVED');
+				$this->success('ACP_AD_SETTINGS_SAVED');
+			}
+			else
```
VSEphpbb: the else is not needed (since the if condition above terminates with a trigger error.

migrations/v10x/m6_hide_for_group.php
```diff
@@ -17,7 +17,7 @@ class m6_hide_for_group extends \phpbb\db\migration\migration
 	*/
 	public function effectively_installed()
 	{
-		return isset($this->config['phpbb_admanagement_hide_groups']);
+		return $config_text->get('phpbb_admanagement_hide_groups') !== null;
```
VSEphpbb: $config_text is not defined yet. You need to get it from the container. TO use the container this migration needs to actually extend the `container_aware_migration` class instead of the usual `migration` class.

controller/admin_controller.php
senky: It seems easier to read this way to me, but OK.

senky:
!unset missing tests

migrations/v10x/m6_hide_for_group.php
VSEphpbb: Actually the way to do this is to extend `\container_aware_migration`.

Then:
```
$config_text = $this->container->get('config_text');
```

controller/admin_controller.php
VSEphpbb: I agree it can be easier to read but I've become a victim of chasing metrics, and Scrutinizer, Code Climate and even my IDE don't like using an else clause, unless absolutely necessary.

controller/admin_controller.php
senky: Oh you are right. It's other way around. After you add a user to a group that has ads hidden, it waits for a cache destroy to take effect. In that case, we can simply ignore the time lag, I guess...

PR #21, opened 2017-06-13
Some fixes and cleanup (mostly in tests)

VSEphpbb:
@senky 

just some clean up based on issues my IDE discovered


tests/event/setup_ads_test.php
```diff
@@ -30,6 +22,8 @@ public function data_setup_ads()
 		return array(
 			array(array(1)),
 			array(array(2)),
+			array(array(1, 2)),
```
senky: I don't understand this and next line additions. They are not used in test.

VSEphpbb: Just adding more diversity. Since in reality, multiple groups may be selected, not just 1 group. So updated tests to make sure it works for users with multiple groups.

tests/functional/end_date_test.php
```diff
@@ -46,17 +46,26 @@ public function test_future_end_date_displays()
...
 
-	public function test_past_end_date_is_not_displayed()
+	/**
+	 * @depends test_future_end_date_displays
+	 */
+	public function test_past_end_date_is_not_displayed($ad_code)
 	{
-		$ad_code = $this->create_ad('below_footer', '2000-01-01');
+		// Change the ads end date to a time long ago
+		$sql = 'UPDATE phpbb_ads
```
senky: Since you change end date with SQL query here, second parameter to `create_ad` is not used anymore and should be removed as well.

VSEphpbb: It is still used in `test_future_end_date_displays`

I had to change it in SQL here because the original test was not working (because you're not allowed to create an ad with an already expired date)

senky: Oh right. Great catch!

PR #28, opened 2017-06-14
Some small fixes

VSEphpbb:


VSEphpbb:
@senky please review these updates

language/en/acp.php
```diff
@@ -19,66 +19,68 @@
 }
 
 $lang = array_merge($lang, array(
+	// Manage ads
 	'AD_NAME'				=> 'Name',
-	'AD_NAME_EXPLAIN'		=> 'This is only used for your recognition of this advertisement.',
+	'AD_NAME_EXPLAIN'		=> 'The ad name is only used to help you identify this advertisement.',
```
senky: I would like to see "advertisement" being used instead of "ad" everywhere. No real reason here, I just feel it's more appropriate. But you are the native speaker, so it's your call.

VSEphpbb: I was feeling the opposite just because advertisement is a mouthful. But I also think ad can be a little too short sometimes. I think there's a good balance of using "ad" after we have established "advertisement" first...ie: use the short version in subsequent usages after the long version, when used in sentences. However here I think we don't need to say "ad" at the beginning of this sentence at all.

senky:
Tests fail, because you added `U_ACTION` to `setup()` method. In `admin_controller_test::test_mode_manage`, replace:
```
		$this->template->expects($this->once())
			->method('assign_var')
			->with('S_PHPBB_ADS', true);
```
with:
```
		$this->template->expects($this->once())
			->method('assign_vars')
			->with(array(
				'S_PHPBB_ADS'	=> true,
				'U_ACTION'		=> $this->u_action,
			));
```

language/en/acp.php
senky: I like your proposition. So remove it here. But I would like to see "advertisement" below, in `CONFIGURE_AD`.

PR #30, opened 2017-06-16
Make sure ads use full width of enclosing div

### LINKED ISSUE #29: Centering issues again
@senky @Derky 

Since we added an inner div to our ad templates, we have centering issues again.

This time the issue affects users who want to add like an Advertisement heading to their ads.

screen shot 2017-06-16 at 2 14 58 pm

We can make our inner div have a width of 100%. But that introduces a new problem, because the flex-box is centering the inner div, which would now be full width, so any ad codes inside will be left justified again.

screen shot 2017-06-16 at 2 51 26 pm

This then forces the user to find their own way to center the ad in their ad code.

Basically what we need is full width support of the page (i.e.: the parent div the ad is in) and the ad needs to be centered, without applying any of its formatting to the internals of the ad itself.

VSEphpbb:
Fixes #29 

@senky can you test and see what you think

### DIFF
--- a/adm/style/manage_ads.html
+++ b/adm/style/manage_ads.html
@@ -11,9 +11,8 @@
 	{% if PREVIEW %}
 		<fieldset>
 			<legend>{{ lang('PREVIEW') }}</legend>
-				<div class="phpbb-ads-center">
-					{{ PREVIEW }}
-				</div>
+			{% set PHPBB_ADS_CODE = PREVIEW %}
+			{% include '@phpbb_ads/phpbb_ads_default.html' %}
 		</fieldset>
 	{% endif %}
 
--- a/adm/style/manage_ads.html
+++ b/adm/style/manage_ads.html
@@ -12,9 +12,7 @@
 		<fieldset>
 			<legend>{{ lang('PREVIEW') }}</legend>
 				<div class="phpbb-ads-center">
-					<div class="phpbb-ads-inner">
-						{{ PREVIEW }}
-					</div>
+					{{ PREVIEW }}
 				</div>
 		</fieldset>
 	{% endif %}
--- a/adm/style/manage_ads.html
+++ b/adm/style/manage_ads.html
@@ -12,7 +12,9 @@
 		<fieldset>
 			<legend>{{ lang('PREVIEW') }}</legend>
 				<div class="phpbb-ads-center">
-					<div>{{ PREVIEW }}</div>
+					<div class="phpbb-ads-inner">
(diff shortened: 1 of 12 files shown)


VSEphpbb:
@senky what do you think of putting our AD template into a single template file and doing this in each template event:

phpbb_ads_default.html:
```html
<div class="phpbb-ads-center" style="{{ PHPBB_ADS_STYLE }}">
	{{ PHPBB_ADS_CODE }}
</div>
```

And then in the temp. events, something like:
```twig
{% if AD_AFTER_PROFILE %}
	{% set PHPBB_ADS_STYLE = 'margin: 10px 0;' %}
	{% set PHPBB_ADS_CODE = AD_AFTER_PROFILE %}
	{% include 'phpbb_ads_default.html' %}
{% endif %}
```
The advanatage to this is anytime we want to adjust our ad template, we only have to do it in the one file instead of all of them.

senky:
@VSEphpbb so you simply reverted #11 ? This looks like the best solution to me. It centers everything correctly except for bare text, which can be simply solved by wrapping ad code by additional `div`.

Default ad template looks good to me. Will you add it to this PR, or should I create a new one?

VSEphpbb:
I added the new logic / default template here.

And yes I reverted our inner div tag as it caused more problems than solutions.

So in general, most ad codes pasted into this ext will just work now.  In some few case, ie text only with links, we just need to recommend they wrap the affected code inside a div.

senky:
I am thinking about such improvement:

`phpbb_ads_default.html`:
```
{% if PHPBB_ADS_CODE %}
<div class="phpbb-ads-center" style="{{ PHPBB_ADS_STYLE }}">
	{{ PHPBB_ADS_CODE }}
</div>
{% endif %}
```

other event files:
```
{% set PHPBB_ADS_STYLE = 'margin: 10px 0;' %}
{% set PHPBB_ADS_CODE = AD_AFTER_PROFILE %}
{% include 'phpbb_ads_default.html' %}
```

senky:
Or maybe even better:

```
{% if PHPBB_ADS_CODE %}
<div class="phpbb-ads-center" style="{{ PHPBB_ADS_STYLE }}">
	{{ PHPBB_ADS_CODE }}
</div>
{% endif %}
```

```
{% include 'phpbb_ads_default.html' with {'PHPBB_ADS_CODE': AD_AFTER_PROFILE, 'PHPBB_ADS_STYLE': 'margin: 10px 0;' } %}
```

PR #31, opened 2017-06-18
AdBlocker friendly message

senky:
Admin can decide on displaying friendly message on locations where AdBlocker hides advertisements.

https://trello.com/c/L4D5Bwvx/7-adblocker-friendly-message

### DIFF
--- a/event/main_listener.php
+++ b/event/main_listener.php
@@ -29,6 +29,9 @@ class main_listener implements EventSubscriberInterface
 	/** @var \phpbb\config\db_text */
 	protected $config_text;
 
+	/** @var \phpbb\config\config */
+	protected $config;
+
 	/** @var \phpbb\ads\ad\manager */
 	protected $manager;
 
@@ -52,14 +55,16 @@ class main_listener implements EventSubscriberInterface
 	* @param \phpbb\template\template				$template			Template object
 	* @param \phpbb\user							$user				User object
 	* @param \phpbb\config\db_text					$config_text		Config text object
-	* @param \phpbb\ads\ad\manager			$manager			Advertisement manager object
-	* @param \phpbb\ads\location\manager	$location_manager	Template location manager object
+	* @param \phpbb\config\config					$config				Config object
+	* @param \phpbb\ads\ad\manager					$manager			Advertisement manager object
+	* @param \phpbb\ads\location\manager			$location_manager	Template location manager object
 	*/
-	public function __construct(\phpbb\template\template $template, \phpbb\user $user, \phpbb\config\db_text $config_text, \phpbb\ads\ad\manager $manager, \phpbb\ads\location\manager $location_manager)
+	public function __construct(\phpbb\template\template $template, \phpbb\user $user, \phpbb\config\db_text $config_text, \phpbb\config\config $config, \phpbb\ads\ad\manager $manager, \phpbb\ads\location\manager $location_manager)
 	{
 		$this->template = $template;
 		$this->user = $user;
 		$this->config_text = $config_text;
+		$this->config = $config;
 		$this->manager = $manager;
 		$this->location_manager = $location_manager;
 	}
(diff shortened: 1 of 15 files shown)


senky:
!set missing tests

senky:
!set WIP

language/en/common.php
```diff
@@ -0,0 +1,24 @@
...
+	exit;
+}
+
+if (empty($lang) || !is_array($lang))
+{
+	$lang = array();
+}
+
+$lang = array_merge($lang, array(
+	'ADBLOCKER_TITLE'	=> 'Adblocker detected',
```
VSEphpbb: Ad blocker instead of Adblocker ?

senky:
!unset WIP

senky:
!unset missing tests

event/main_listener.php
```diff
@@ -100,5 +105,11 @@ public function setup_ads()
 				));
 			}
 		}
+
+		// Display Ad blocker friendly message if allowed
+		if ($this->config['phpbb_ads_adblocker_message'])
+		{
+			$this->template->assign_var('S_DISPLAY_ADBLOCKER', true);
```
VSEphpbb: I think we can eliminate the conditional and just do:
```
$this->template->assign_var('S_DISPLAY_ADBLOCKER', $this->config['phpbb_ads_adblocker_message']);
```

language/en/acp.php
```diff
@@ -79,8 +79,10 @@
 	'AD_AFTER_PROFILE_DESC'			=> 'Displays after member profile page content.',
 
 	// Settings
-	'HIDE_GROUPS'			=> 'Hide advertisements for groups',
-	'HIDE_GROUPS_EXPLAIN'	=> 'Members of selected groups will not see any advertisement. Use CTRL+CLICK (or CMD+CLICK on Mac) to select/deselect more than one group.',
+	'ADBLOCKER_MESSAGE'				=> 'Ad blocker friendly message',
+	'ADBLOCKER_MESSAGE_EXPLAIN'		=> 'Display friendly message to users who use ad blocker.',
```
VSEphpbb: I think it could say more like:
> Display a polite message to visitors using ad blockers, advising them to consider disabling ad blocking on this forum.

language/en/common.php
```diff
@@ -19,6 +19,6 @@
 }
 
 $lang = array_merge($lang, array(
-	'ADBLOCKER_TITLE'	=> 'Adblocker detected',
+	'ADBLOCKER_TITLE'	=> 'Ad blocker detected',
 	'ADBLOCKER_MESSAGE'	=> 'Our website is made possible by displaying online advertisements to our visitors. Please consider supporting us by disabling your ad blocker on our website.',
```
VSEphpbb: This is a good line but I know people will eventually ask us to let them customize this themselves. What I think we should do, but later (during final polishing) is make this an ACP option using config_text. But we keep this as our default. So you would insert this lang key by default into the ACP field (sort of like how naming permission roles works). We would always run whatever is in this field through the lang() so that if it's our key, it uses the correct translation of our default text here, otherwise it will display whatever they wrote in custom.

migrations/v10x/m7_adblocker.php
```diff
@@ -0,0 +1,42 @@
...
+namespace phpbb\ads\migrations\v10x;
+
+class m7_adblocker extends \phpbb\db\migration\migration
+{
+	/**
+	* {@inheritDoc}
+	*/
+	public function effectively_installed()
+	{
+		return isset($this->config['phpbb_ads_adblocker_message']);
```
VSEphpbb: this line is fine, but fyi, I think it's kinda cool to also use this (you don't have to change this, it's just FYI for you):

```php
return $this->config->offsetExists('phpbb_ads_adblocker_message');
```

language/en/common.php
senky: Well, board may be multi-lingual. If you only allow admin to insert plain text or lang key in the field, you still kind of force them to use lang files. They can simply rewrite desc in the file. So either proper solution (like with custom profile fields), or none. Nevertheless, it is passé for now. 🙂

VSEphpbb: That's the price to pay. If board is multi-lingual they can use the default key, or know that they are writing their own text in just the board's primary language. There's stuff in phpBB already that work with that caveat (take custom BBCode help lines, for example).

A smarter admin would of course, use their own key and add it to their board themselves, either by editing a core lang file or writing a simple extension.

But as you say, this is just something to add to the requests wish list.

PR #32, opened 2017-06-19
Ad priority

senky:


senky:
@VSEphpbb @Derky 2 things:

1. Maybe we should display `5` in ad priority input when user adds new ad?

VSEphpbb:
Hmm. So a little more info here. I have assigned 2 ads to the same location. The get_ads() function is returning both of them in an array.
```
array
  'location_id' => 'above_header'
  'ad_code' => 'Low Priority'
array
  'location_id' => 'above_header'
  'ad_code' => 'High Priority'
```
When this array is used in setup_ads(), the last array item always overwrites the previous array items for output. In this case, because of the ORDER BY ad_priority, the High Priority ad is always being shown.

There is no mechanism in place that I can see for looking at these 2 ads priority number, and randomly choosing 1 of them, giving preference (or weighting) to the higher number over the lower number.

Also now that we require a priority of some value for all ads, is the RAND() in our SQL is neccessary anymore??

I did a quick google search and while I recommend you come up with the solution you think is the best @senky ... wether that is in PHP or SQL ... I did find this simple solution https://stackoverflow.com/questions/7155924/how-to-query-random-records-in-a-mysql-table-while-factoring-in-a-priority-syste:

```sql
ORDER BY (RAND() * a.ad_priority)
```

Also I noticed that we should probably not functional test for this, because in reality the ad can be either the high or low ad, so in this case, your current functional test would fail sometimes.

Derky:
@senky You mean use 5 as default priority instead of leaving it blank? That's fine yeah.

senky:
@VSEphpbb so this is how it works. You get back more than one result from the query because I followed someone's advise on how to group by ad location but still return ad code (which is not aggregated). So the only solution seemed to be the current one.

Now, I thought about applying priority to the query, and I decided to always favour ad with higher priority. I suppose most ads will have more than one location, so priority might hide it for one location, but still display on the another. Another reason is index keys - query is fast that way.

But your proposed solution looks good as well. Query will be slower probably, but will rotate ads better and still use priority as reference point.

Regarding functional tests, as soon as I apply your solution, the test will be obsolete, of course. But I can imagine one - load index 100x and compare, that ad with higher priority was displayed more times. In as high number as 100, it must be always true. But such test would take tremendous amount of time. Probably not worth it.

VSEphpbb:
Regarding performance, the speed seems to improve with the proposed query solution.

I did a test and the original query:
0.0032s

With the new query I found:
0.0029s

adm/style/manage_ads.html
```diff
@@ -41,6 +41,10 @@
 				<dd><input class="text" type="text" id="ad_end_date" name="ad_end_date" value="{{ AD_END_DATE }}" size="20" maxlength="20" /></dd>
 			</dl>
 			<dl>
+				<dt><label for="ad_priority">{{ lang('AD_PRIORITY') ~ lang('COLON') }}</label><br /><span>{{ lang('AD_PRIORITY_EXPLAIN') }}</span></dt>
```
VSEphpbb: I would prefer this setting be relocated to above/below the ad locations since the priority is mostly related to the locations.

VSEphpbb: Also when creating a new ad, we should pre-fill this field with the default value of 5.

controller/admin_controller.php
```diff
@@ -209,6 +209,7 @@ public function action_add()
 		else
 		{
 			$this->assign_locations();
+			$this->template->assign_var('AD_PRIORITY', 5);
```
VSEphpbb: Hmm, I don't really like this, just because it feels so specific and out of place.

Let's try this:
First, we should make 5 a class consatnt since it's used more than once, like:
```php
const DEFAULT_PRIORITY = 5;
```
Then lets instead of adding the value to our form, let's use a form placeholder instead? This can be done directly in the template thanks to twig:
```twig
placeholder="{{ constant('\\phpbb\\ads\\controller\\admin_controller::DEFAULT_PRIORITY') }}"
```

language/en/acp.php
```diff
@@ -46,6 +48,7 @@
 	'AD_NAME_REQUIRED'			=> 'Name is required.',
 	'AD_NAME_TOO_LONG'			=> 'Name length is limited to %d characters.',
 	'AD_END_DATE_INVALID'		=> 'The end date was invalid or has already expired.',
+	'AD_PRIORITY_INVALID'		=> 'The priority is invalid. Please set a number between 1 and 10.',
```
VSEphpbb: Lets keep the same format as our other validation errors. We should change "was' to "is" in AD_END_DATE_INVALID

controller/admin_controller.php
```diff
@@ -443,6 +444,7 @@ protected function get_form_data($form_name)
 			'ad_enabled'	=> $this->request->variable('ad_enabled', 0),
 			'ad_locations'	=> $this->request->variable('ad_locations', array('')),
 			'ad_end_date'	=> $this->request->variable('ad_end_date', ''),
+			'ad_priority'	=> $this->request->variable('ad_priority', 5),
```
VSEphpbb: Can you use the priority constant here too

PR #35, opened 2017-06-21
#34: Optimize functional tests

### LINKED ISSUE #34: Optimize functional tests
- Add `disable_all_ads()` to `functional_base`, but use direct SQL query instead of crawler in loop.
- Remove third parameter from `create_ad` in `functional_base`
- Remove dependency from `test_adblocker_code_is_present`, because it depends on test of different class
- Remove difficult readable dependency from `test_past_end_date_is_not_displayed` - simply create another ad, update it's end date and test it's not present
- Rationalize use of `reset_groups()` in `hide_group_test` - it needs to be called after every test, so use `tearDown()`
- Split `acp_test`, because it is too big
- Refactor `acp_test` (introduce `get_manage_page()`, `get_settings_page()` and `submit_with_error($error_lang)` (maybe we won't need to split the class after that)

senky:
Resolves #34 

Checklist: 
- [x] Add disable_all_ads() to functional_base, but use direct SQL query instead of crawler in loop.
- [x] Remove dependency from test_adblocker_code_is_present, because it depends on test of different class
- [x] Rationalize use of reset_groups() in hide_group_test - it needs to be called after every test, so use tearDown()
- [x] Split acp_test, because it is too big
- [x] Refactor acp_test (introduce get_manage_page(), get_settings_page() and submit_with_error($error_lang) (maybe we won't need to split the class after that)

### DIFF
--- a/tests/functional/end_date_test.php
+++ b/tests/functional/end_date_test.php
@@ -43,24 +43,20 @@ class end_date_test extends functional_base
 
 		// Confirm above header ad is present
 		$this->assertContains($ad_code, $crawler->html());
-
-		return $ad_code;
 	}
 
-	/**
-	 * @depends test_future_end_date_displays
-	 */
-	public function test_past_end_date_is_not_displayed($ad_code)
+	public function test_past_end_date_is_not_displayed()
 	{
+		$ad_code = $this->create_ad('below_header', '2034-01-01');
+
 		// Change the ads end date to a time long ago
 		$sql = 'UPDATE phpbb_ads
-			SET ad_end_date = ' . strtotime('2000-01-01') . '
-			WHERE ad_end_date = ' . strtotime('2035-01-01');
+			SET ad_end_date = ' . strtotime('2000-01-01');
 		$this->db->sql_query($sql);
 
 		$crawler = self::request('GET', 'index.php');
 
-		// Confirm above header ad is present
+		// Confirm below header ad is not present
 		$this->assertNotContains($ad_code, $crawler->html());
 	}
 }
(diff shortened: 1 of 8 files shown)


senky:
@VSEphpbb @Derky what do you think about splitting acp_test?

Derky:
Yeah splitting manage and settings makes sense.

tests/functional/end_date_test.php
```diff
@@ -50,24 +43,20 @@ public function test_future_end_date_displays()
...
-		return $ad_code;
 	}
 
-	/**
-	 * @depends test_future_end_date_displays
-	 */
-	public function test_past_end_date_is_not_displayed($ad_code)
+	public function test_past_end_date_is_not_displayed()
 	{
+		$ad_code = $this->create_ad('below_header', '2034-01-01');
```
VSEphpbb: Why not just make this test depend on `test_no_end_date_displays` and use the SQL to add an expired end date to it, rather than creating a new ad and messing around with the date like this?

senky: I personally don't like test dependencies. IMO, every test should be able to run separately.

I especially didn't like this test's dependency, which has taken dependency's returned values as argument. This simple line decouples the tests.

VSEphpbb: You should get rid of the date in the create_ad in this test though. It serves no purpose at all in this one, since you are really adding the correct date to be tested with the following SQL.

FYI passing a result from one test to another test is a useful feature. In this case, another ad was not really needed since this test is really just updating an ad's date and then re-testing it. It didn't really need to build a whole new ad for that and could quite easily use either of the ads created in the prior 2 tests.

senky: Reverted then

PR #37, opened 2017-06-22
Improve and fix ad priority

VSEphpbb:
What do you think of this test for really testing that priority works?

### DIFF
--- a/ad/manager.php
+++ b/ad/manager.php
@@ -262,6 +262,7 @@ class manager
 
 	/**
 	* Get the random statement for this database layer
+	* Random function should generate a float value between 0 and 1
 	*
 	* @return	string	Random statement for current database layer
 	*/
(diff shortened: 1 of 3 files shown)


VSEphpbb:
Hmm. We may have a problem on SQLite

senky:
1. I think you can increase iterations to 100. It will drastically lower chance of skipping and querying DB 100 times is a matter of a second.

Let me dive into SQLite and see what's the problem here. For now, it looks like ad priority is ignored in the query with this DB type.

ad/manager.php
```diff
@@ -270,10 +271,15 @@ protected function sql_random()
...
+				return 'VALUE()';
+
 			case 'postgres':
+				return 'RANDOM()';
+
+			// https://stackoverflow.com/a/35369410/2908600
 			case 'sqlite':
 			case 'sqlite3':
-				return 'RANDOM()';
+				return '(0.5 - RANDOM() / CAST(-9223372036854775808 AS REAL) / 2)';
```
senky: What is the logic behind this? It's range is from -1.5 to 2.5.

I really think, that since range 0 to 1 works everywhere else, it must work here as well. I assume this should work perfectly:
`ABS(RANDOM() / CAST(-9223372036854775808 AS REAL))`

VSEphpbb: The range on this is 0 to 1. The `0.5 -` at the top ensures that.

VSEphpbb: https://stackoverflow.com/questions/23785143/need-help-adding-random-float-in-sqlite/35369410#35369410

PR #42, opened 2017-07-11
#33: Tests should be refactored (eventually)

### LINKED ISSUE #33: Tests should be refactored (eventually)
Not a high priority but in my opinion our tests of the admin controller should be refactored.

Currently we are testing several important classes and methods (the ad manager and location managers) as a side effect of testing the admin controller and event listeners.

This is causing us to miss some things.

What we should have is unit testing of the ad manager's methods, testing each method with both expected and unexpected data. Same for the location manager.

Then in our admin controller and event listener, since they are not really processing data so much as just passing data to the managers, we should test the controller/listener by using stubs/mocks of the managers and just checking that they are calling the expected methods and handing them the expected data.

This way we are directly testing our work-horses (the managers) and for proper data handling and results, and testing our "controllers" for using the managers correctly.

senky:
Closes #33 

- [x] Ad manager
- [x] Location manager
- [x] Replace classes with mocks in controller/listener tests

### DIFF
--- a/tests/ad/delete_ad_locations_test.php
+++ b/tests/ad/delete_ad_locations_test.php
@@ -1,0 +1,33 @@
+<?php
+/**
+ *
+ * Advertisement management. An extension for the phpBB Forum Software package.
+ *
+ * @copyright (c) 2017 phpBB Limited <https://www.phpbb.com>
+ * @license GNU General Public License, version 2 (GPL-2.0)
+ *
+ */
+
+namespace phpbb\ads\tests\ad;
+
+class delete_ad_locations_test extends ad_base
+{
+	/**
+	 * Test delete_ad_locations() method
+	 */
+	public function test_delete_ad_locations()
+	{
+		$manager = $this->get_manager();
+
+		$manager->delete_ad_locations(6);
+
+		$sql = 'SELECT location_id
+			FROM phpbb_ad_locations
+			WHERE ad_id = 6';
+		$result = $this->db->sql_query($sql);
+		$affected_rows = $this->db->sql_affectedrows();
+		$this->db->sql_freeresult($result);
+
+		$this->assertEquals(0, $affected_rows);
+	}
+}
--- a/tests/ad/delete_ad_locations_test.php
+++ b/tests/ad/delete_ad_locations_test.php
@@ -19,15 +19,10 @@ class delete_ad_locations_test extends ad_base
 	{
 		$manager = $this->get_manager();
 
+		$this->assertNotEmpty($manager->get_ad_locations(6));
+
 		$manager->delete_ad_locations(6);
 
-		$sql = 'SELECT location_id
-			FROM phpbb_ad_locations
-			WHERE ad_id = 6';
-		$result = $this->db->sql_query($sql);
-		$affected_rows = $this->db->sql_affectedrows();
-		$this->db->sql_freeresult($result);
-
-		$this->assertEquals(0, $affected_rows);
+		$this->assertEmpty($manager->get_ad_locations(6));
 	}
 }
--- a/tests/ad/delete_ad_locations_test.php
+++ b/tests/ad/delete_ad_locations_test.php
@@ -26,7 +26,6 @@ class delete_ad_locations_test extends ad_base
 		$this->assertEmpty($manager->get_ad_locations(6));
 	}
 
-
 	/**
 	 * Test delete_ad_locations() method with non-existent ad
 	 */
--- a/tests/ad/delete_ad_locations_test.php
+++ b/tests/ad/delete_ad_locations_test.php
@@ -25,4 +25,27 @@ class delete_ad_locations_test extends ad_base
 
 		$this->assertEmpty($manager->get_ad_locations(6));
 	}
+
+
+	/**
+	 * Test delete_ad_locations() method with non-existent ad
+	 */
+	public function test_delete_ad_locations_no_ad()
+	{
+		$manager = $this->get_manager();
+
+		$sql = 'SELECT COUNT(ad_id) as total_ad_locations
+			FROM phpbb_ad_locations';
+		$result = $this->db->sql_query($sql);
+		$total_ad_locations = $this->db->sql_fetchfield('total_ad_locations');
+		$this->db->sql_freeresult($result);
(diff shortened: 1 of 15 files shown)


senky:
!set WIP

senky:
!unset WIP

senky:
Why, SQLite, why?
Why?

tests/ad/delete_ad_locations_test.php
```diff
@@ -0,0 +1,33 @@
...
+	/**
+	 * Test delete_ad_locations() method
+	 */
+	public function test_delete_ad_locations()
+	{
+		$manager = $this->get_manager();
+
+		$manager->delete_ad_locations(6);
+
+		$sql = 'SELECT location_id
```
VSEphpbb: Instead of a query, do 
```
$this->assertEmpty($manager->get_ad_locations(6));
```

Also before you delete the ad, assert it exists first:
```
$this->assertNotEmpty($manager->get_ad_locations(6));
```

tests/ad/delete_ad_test.php
```diff
@@ -0,0 +1,35 @@
...
+class delete_ad_test extends ad_base
+{
+	/**
+	 * Test delete_ad() method
+	 */
+	public function test_delete_ad()
+	{
+		$manager = $this->get_manager();
+
+		$affected_rows = $manager->delete_ad(6);
```
VSEphpbb: Same as above here. Test ad exists first with `get_ad(6)` and after deleting assert `get_ad(6)` is false.

tests/ad/get_all_ads_test.php
```diff
@@ -0,0 +1,63 @@
...
+	/**
+	 * Test get_all_ads() method
+	 */
+	public function test_get_all_ads()
+	{
+		$manager = $this->get_manager();
+
+		$ads = $manager->get_all_ads();
+
+		$this->assertEquals(array(
```
VSEphpbb: This test could be annoying if we make changes to the fixture. So i would simplify it to just check one of the array columns... Probably IDs for simplicity, i.e.:

```php
$ads = $manager->get_all_ads();
$ad_ids = array_column($ads, 'ad_id');
$this->assertEquals(array(1,2,3,4,5,6), $ad_ids);
```

tests/ad/insert_ad_locations_test.php
```diff
@@ -0,0 +1,71 @@
...
+	 *
+	 * @dataProvider insert_ad_locations_data
+	 */
+	public function test_insert_ad_locations($ad_id, $ad_locations, $expected)
+	{
+		$manager = $this->get_manager();
+
+		$manager->insert_ad_locations($ad_id, $ad_locations);
+
+		$sql = 'SELECT location_id
```
VSEphpbb: Also here, inseatd of SQL (we want to avoid queries as much as possible and instead rely on our methods);

use `$manager->get_ad_locations($ad_id);` to get the actual array of locations and assert it equals the $expected array

tests/ad/insert_ad_test.php
```diff
@@ -0,0 +1,69 @@
...
+	 *
+	 * @dataProvider insert_ad_data
+	 */
+	public function test_insert_ad($data)
+	{
+		$manager = $this->get_manager();
+
+		$ad_id = $manager->insert_ad($data);
+
+		$this->assertGreaterThan(0, $ad_id);
```
VSEphpbb: Assert greater than `6` to be sure we're adding new ads not already in our fixture.

tests/ad/insert_ad_test.php
```diff
@@ -0,0 +1,69 @@
...
+
+		$sql = 'SELECT ad_name
```
VSEphpbb: Also here, instead of query use `manager->get_ad($ad_id)` and assert it matches the $data array we just inserted.

tests/ad/load_memberships_test.php
```diff
@@ -0,0 +1,29 @@
...
+class load_memberships_test extends ad_base
+{
+	/**
+	 * Test load_memberships() method
+	 */
+	public function test_load_memberships()
+	{
+		$manager = $this->get_manager();
+
+		$memberships = $manager->load_memberships(1);
```
VSEphpbb: Use a test data provider to test this method with good data and bad data. i.e.:

// user id, expected
1, array(1,3)
2, array()
null, array()
0, array()

tests/ad/update_ad_test.php
```diff
@@ -0,0 +1,80 @@
...
+	 */
+	public function test_update_ad($ad_id, $data, $affected_rows)
+	{
+		$manager = $this->get_manager();
+
+		$updated = $manager->update_ad($ad_id, $data);
+
+		$this->assertEquals($affected_rows, $updated);
+
+		$sql = 'SELECT ' . implode(', ', array_keys($data)) . '
```
VSEphpbb: Also again, preferable to use `get_ad($ad_id)` and check that the array values from the $data match the one's now in the ad. i.e.:

```php
$ad = $manager->get_ad($ad_id);
foreach ($data as $key => $value)
{
    $this-assertEquals($value, $ad[$key])
}
```
No need for affected rows then too.

tests/ad/delete_ad_locations_test.php
VSEphpbb: It is also good to have a separate test in this file to test when trying to delete a non existent ad, like deleting ad_id 0. You can test that by then getting all ads and asserting the count of the array is still 6

tests/ad/delete_ad_locations_test.php
```diff
@@ -25,4 +25,27 @@ public function test_delete_ad_locations()
...
+		$this->db->sql_freeresult($result);
+
+		$manager->delete_ad_locations(0);
+
+		$sql = 'SELECT COUNT(ad_id) as total_ad_locations
+			FROM phpbb_ad_locations';
+		$result = $this->db->sql_query($sql);
+		$this->assertEquals($this->db->sql_fetchfield('total_ad_locations'), $total_ad_locations);
+		$this->db->sql_freeresult($result);
+	}
```
VSEphpbb: Also a similar thing for the delete ad test

PR #43, opened 2017-07-11
Views & Clicks

senky:
Extension will hold statistics of views/clicks/view duration for each advertisement. Admin can set views/clicks limit after which advertisement will be disabled. Registered users’ views and clicks will be logged for better insights.


senky:
!set missing tests

senky:
I would like to see #42 merged before I fix/write tests for this one.

event/main_listener.php
```diff
@@ -97,16 +102,26 @@ public function setup_ads()
...
 			}
+
+			$this->manager->increment_ads_views($ad_ids);
 		}
 
 		// Display Ad blocker friendly message if allowed
-		$this->template->assign_var('S_DISPLAY_ADBLOCKER', $this->config['phpbb_ads_adblocker_message']);
+		$this->template->assign_vars(array(
+			'S_DISPLAY_ADBLOCKER'	=> $this->config['phpbb_ads_adblocker_message'],
+			'U_PHPBB_ADS_CLICK'		=> $this->controller_helper->route('phpbb_ads_click', array('ad_id' => 0)),
```
Derky: Please use `UA_` for JavaScript URLs. (See "Variables" in the [Coding Guidelines](https://area51.phpbb.com/docs/32x/coding-guidelines.html#templates))

ad/manager.php
```diff
@@ -98,6 +102,40 @@ public function get_all_ads()
...
+	* Increment clicks for specified ad
+	*
+	* @param	array	$ad_id	ID of an ad to increment clicks
+	* @return	void
+	*/
+	public function increment_ad_clicks($ad_id)
+	{
+		$sql = 'UPDATE ' . $this->ads_table . '
+			SET ad_clicks = ad_clicks + 1
+			WHERE ad_id = ' . $ad_id;
```
Derky: Please cast `$ad_id` to (int)

(yes I saw the routing requirement 😛)

event/main_listener.php
```diff
@@ -97,16 +102,37 @@ public function setup_ads()
 		if (!array_intersect($user_groups, $hide_groups))
 		{
 			$location_ids = $this->location_manager->get_all_location_ids();
+			$ad_ids = array();
 
 			foreach ($this->manager->get_ads($location_ids) as $row)
 			{
+				$ad_ids[] = $row['ad_id'];
+
+				if ($this->config['phpbb_ads_enable_clicks'])
```
VSEphpbb: Actually we should not be doing this {COUNT_CLICKS} thing like this. We should add it ourselves, and not tell the user they have to add a COUNT_CLICKS placeholder to their code.

We can just do this in our ad templates by inserting `{% if S_PHPBB_ADS_ENABLE_CLICKS %}data-ads-click="{{ AD_ID }}"{% endif %}` ourselves.

event/main_listener.php
```diff
@@ -97,16 +102,37 @@ public function setup_ads()
...
+			}
 		}
 
 		// Display Ad blocker friendly message if allowed
-		$this->template->assign_var('S_DISPLAY_ADBLOCKER', $this->config['phpbb_ads_adblocker_message']);
+		$this->template->assign_vars(array(
+			'S_DISPLAY_ADBLOCKER'		=> $this->config['phpbb_ads_adblocker_message'],
+			'UA_PHPBB_ADS_CLICK'		=> $this->controller_helper->route('phpbb_ads_click', array('ad_id' => 0)),
+			'S_PHPBB_ADS_ENABLE_CLICKS'	=> (bool) $this->config['phpbb_ads_enable_clicks'],
+		));
```
VSEphpbb: I might actually do this instead:
```php
// Display Ad blocker friendly message if allowed
$this->template->assign_var('S_DISPLAY_ADBLOCKER', $this->config['phpbb_ads_adblocker_message']);

// Add click tracking template variables
if ($this->config['phpbb_ads_enable_clicks'])
{
		'UA_PHPBB_ADS_CLICK'		=> $this->controller_helper->route('phpbb_ads_click', array('ad_id' => 0)),
		'S_PHPBB_ADS_ENABLE_CLICKS'	=> true,
	));
}
```

migrations/v10x/m9_views_clicks.php
```diff
@@ -0,0 +1,99 @@
...
+
+	/**
+	 * Add phpbb_ads_enable_views and phpbb_ads_enable_clicks config
+	 *
+	 * @return array Array of data update instructions
+	 */
+	public function update_data()
+	{
+		return array(
+			array('config.add', array('phpbb_ads_enable_views', 1)),
```
VSEphpbb: These should default to 0 please. We don't want to force a feature upon people 😄

styles/all/template/js/clicks.js
```diff
@@ -0,0 +1,7 @@
+$(function() {
+	'use strict';
+
+	$('[data-ads-click]').on('click', function(e) {
+		$.get(u_phpbb_ads_click.replace(/0$/, $(this).attr('data-ads-click')));
```
VSEphpbb: Since data-ads-click really holds the ad id, I might rename it to `data-ads-id` instead, semantically makes more sense.

adm/style/settings_ads.html
```diff
@@ -20,6 +20,16 @@
 				<label><input type="radio" class="radio" name="adblocker_message" value="0"{% if not ADBLOCKER_MESSAGE %} checked="checked"{% endif %} /> {{ lang('DISABLED') }}</label></dd>
 		</dl>
 		<dl>
+			<dt><label for="enable_views">{{ lang('ENABLE_VIEWS') ~ lang('COLON') }}</label><br /><span>{{ lang('ENABLE_VIEWS_EXPLAIN') }}</span></dt>
```
senky: I am not sure what you are talking about here. This is settings form, not new ad form.

controller/front_controller.php
```diff
@@ -0,0 +1,48 @@
...
+ * @license GNU General Public License, version 2 (GPL-2.0)
+ *
+ */
+
+namespace phpbb\ads\controller;
+
+/**
+* Front controller
+*/
+class front_controller
```
senky: I thought maybe we will extend it later, so we can group all frontend-related stuff here... But ok, we can rename it later.

adm/style/settings_ads.html
VSEphpbb: I meant on ad add/edit page

controller/front_controller.php
VSEphpbb: The only other front end stuff I think i see coming is for ad owners, which would really warrant a separate controller for that feature.

event/main_listener.php
VSEphpbb: Actually we just have to add this to our default template ;)

senky: I was thinking about this option as well. It is way too simpler for common user, but enabling them to choose where to put onclick event gives them more flexibility One can create a banner, where only a small part is a link to another site. Your solution would count clicks even when user clicks on non-anchored area, which may be undesired solution.

So we should really sort this out. While I agree with your proposition, don't you think user will ask for more flexibility?

event/main_listener.php
```diff
@@ -97,16 +102,37 @@ public function setup_ads()
...
+				$ad_code = htmlspecialchars_decode($ad_code);
 				$this->template->assign_vars(array(
-					'AD_' . strtoupper($row['location_id'])	=> htmlspecialchars_decode($row['ad_code']),
+					'AD_' . strtoupper($row['location_id'])	=> $ad_code,
 				));
 			}
+
+			if ($this->config['phpbb_ads_enable_views'])
+			{
+				$this->manager->increment_ads_views($ad_ids);
```
senky: Nope. `$this->location_manager->get_all_location_ids()` only returns ad locations that `will_display()`. A result is injected into `$this->manager->get_ads($location_ids)`, that returns only ads that will be really displayed. So it works fine.

event/main_listener.php
VSEphpbb: No because we are building this for the purposes of "normal" ads. That means the only thing a user puts in is an actual ad, and the entire thing, whether it's a banner, or text link or both is expected to be a clickable advertisement.

Also, we're providing a fairly rudimentary click tracking thing here. This isn't true analytics, so it will never be perfect. But as far as I'm concerned, if somebody clicks somewhere within the "ad-space" that should count as a click, because for whatever reason a user clicked in the ad-space.

Trying to give this flexibility to users will be a big support issue. Many will not understand it. Will not put it in the right place, for example as written it needs to go inside some sort of HTML element. It's a liability and an easy way for users to do it wrong, and bug us for support over it (and they'll never tell us they did it wrong, they'll just say click tracking is broken).

So we should just implement it ourselves into our defined ad-space in our default template's div tag, and not be concerned with the rare users who may use this feature on ads that are unorthodox.

VSEphpbb: Crap, the unfortunate thing is our div is full screen.

So this deserves another strategy I think. Maybe some JS trickery?

VSEphpbb: Maybe use JS (or php?) to find the anchor in an ad, and replace them with a link to a click counter controller, and that controller counts the click and then redirects the user to the link from the ad?

VSEphpbb: Or see how the Ad management MOD handled this?

event/main_listener.php
VSEphpbb: That's not the result I'm getting. I get 3 ads loaded, but only 1 is actually displayed (in header). The thing to note is the 3 ads being incremented are all assigned to the same location. But because of priority, only 1 of them is displayed.

Before (all 7's)
screen shot 2017-07-12 at 12 06 53 am

Load homepage (1 ad)
screen shot 2017-07-12 at 12 06 59 am

Back to ACP (all 7's incremented to 8's):
screen shot 2017-07-12 at 12 07 11 am

VSEphpbb: Actually the issue was from `$this->manager->get_ads($location_ids)`

VSEphpbb: Basically all the ads assigned to a given location are returned by our query in get_ads...so we need to filter each location down to just 1 ad, and the way things are sorted by our SQL, I think using `array_pop()` to get the last item of each location is what is needed.

ad/manager.php
```diff
@@ -61,28 +66,36 @@ public function get_ad($ad_id)
...
 				ORDER BY (' . $this->sql_random() . ' * a.ad_priority)
 			) z
 			ORDER BY z.location_id';
 		$result = $this->db->sql_query($sql);
 		$data = $this->db->sql_fetchrowset($result);
 		$this->db->sql_freeresult($result);
 
-		return $data;
+		$current_location_id = '';
+		$filtered = array_filter($data, function($row) use (&$current_location_id) {
```
VSEphpbb: I think you can just reuse $data instead of making a new array called $filtered

controller/click_controller.php
```diff
@@ -36,11 +36,16 @@ public function __construct(\phpbb\ads\ad\manager $manager, \phpbb\controller\he
...
-	* @param int	$ad_id	Advertisement ID
-	* @return void
+	* @param	int	$ad_id	Advertisement ID
+	* @return	\Symfony\Component\HttpFoundation\Response	A Symfony Response object
 	*/
 	public function increment_clicks($ad_id)
 	{
+		if (!$ad_id)
+		{
+			throw new \phpbb\exception\http_exception(404, 'NOT_FOUND');
```
VSEphpbb: This needs more work. First if we will throw this exception, NOT_FOUND is a lang key...does it have a translation? Also, does this even apply? Certainly if a user enter a bad url by hand they will get the proper error. What do we want to do if a user entered the URL to a proper ad. Because that is a potential abuse vector, we should definitely have an `is_ajax` type of test here, so that we only process ajax requests and throw another exception for all non-ajax requests (a 403 not auhorised). And finally, if the request is ajax, but the ad id is invalid, how does this handle that with the 404 http_exception above as far as the ajax response goes?

event/main_listener.php
```diff
@@ -107,18 +107,11 @@ public function setup_ads()
...
-					$ad_code = str_replace('{COUNT_CLICKS}', ' data-ads-click="' . $row['ad_id'] . '"', $row['ad_code']);
-				}
-				else
-				{
-					$ad_code = str_replace('{COUNT_CLICKS}', '', $row['ad_code']);
-				}
-				$ad_code = htmlspecialchars_decode($ad_code);
+				$ad_code = $row['ad_code'];
+				$replacement = $this->config['phpbb_ads_enable_clicks'] ? ' data-ads-id="' . $row['ad_id'] . '"' : '';
+				$ad_code = str_replace('{COUNT_CLICKS}', $replacement, $ad_code);
```
VSEphpbb: I figured this out and this really easy to do without the placeholders. We can just modify your ad click JS to delegate the click event to all anchors inside our phpbb-ads div. 

First add the ad id to our templates phpbb-ads div tag in a data attribute:

`data-ad-id="{{ PHPBB_ADS_ID }}"`

Then tweak your click JS to something like:

```javascript
    $(".phpbb-ads").on("click", "a", function (e) {
        e.preventDefault();
        $.get(u_phpbb_ads_click.replace(/0$/, $(this).attr("data-ad-id ")));
        e.click();
    });
```

I have not tested this, but this is the basic idea that should eliminate all this placeholder stuff the user has to bother with.

VSEphpbb: Updated and tested:

```javascript
$("[data-ad-id]").on("click", "a", function (e) {
    $.get(u_phpbb_ads_click.replace(/0$/, $(this).closest("[data-ad-id]").attr("data-ad-id")));
});
```

ad/manager.php
```diff
@@ -98,6 +115,40 @@ public function get_all_ads()
...
+				SET ad_views = ad_views + 1
+				WHERE ' . $this->db->sql_in_set('ad_id', $ad_ids);
+			$this->db->sql_query($sql);
+		}
+	}
+
+	/**
+	* Increment clicks for specified ad
+	*
+	* @param	array	$ad_id	ID of an ad to increment clicks
```
VSEphpbb: int not array

controller/click_controller.php
VSEphpbb: Maybe something like this:

```php
public function increment_clicks($ad_id)
{
	if ($this->request->is_ajax() && !empty($ad_id))
	{
		$this->manager->increment_ad_clicks($ad_id);

		return new \Symfony\Component\HttpFoundation\JsonResponse();
	}

	throw new \phpbb\exception\http_exception(403, 'NOT_AUTHORISED');
}
```

event/main_listener.php
VSEphpbb: We should also wrap this in a jQuery DOM ready function so we make sure the DOM is complete before applying (this may get it working with SCRIPT based ad codes like Google Adsense, but I haven't tested this yet).

senky: Okay. I am not very happy with the code, since anchor is only one way of redirecting user. You can use JS onclick events, etc. But It definitely makes life much easier for admins. So I say, let's go this way and see what the reactions of the people are.

VSEphpbb: We may add `script` along with `a` so any click on a script counts too in our ads...if that works.

controller/click_controller.php
```diff
@@ -21,16 +21,21 @@ class click_controller
...
+	 *
+	 * @param \phpbb\ads\ad\manager    $manager Advertisement manager object
+	 * @param \phpbb\controller\helper $helper  Controller helper object
+	 * @param \phpbb\request\request   $request	Request object
+	 */
+	public function __construct(\phpbb\ads\ad\manager $manager, \phpbb\controller\helper $helper, \phpbb\request\request $request)
 	{
 		$this->manager = $manager;
 		$this->helper = $helper;
+		$this->request = $request;
```
VSEphpbb: Remove helper class

styles/all/template/event/memberlist_view_content_append.html
```diff
@@ -1,3 +1,4 @@
 {% set PHPBB_ADS_STYLE = 'margin: 10px 0;' %}
 {% set PHPBB_ADS_CODE = AD_AFTER_PROFILE %}
+{% set PHPBB_ADS_ID = AD_AFTER_PROFILE_ID %}
```
VSEphpbb: If you would like to combine setting the code and ids into one line / one call of set function that is fine. 

I just wanted the style to be set separate since it is the one customizable line in these templates.

.travis.yml
```diff
@@ -5,7 +5,7 @@ language: php
 
 matrix:
   include:
-    - php: 5.4
+    - php: 5.5
```
VSEphpbb: You also need to add the following changes for EPV:
https://github.com/phpbb-extensions/autogroups/pull/132/files

ad/manager.php
```diff
@@ -61,23 +66,35 @@ public function get_ad($ad_id)
...
 				WHERE a.ad_enabled = 1
 					AND (a.ad_end_date = 0
-						OR a.ad_end_date > ' . time() . ')
-					AND ' . $this->db->sql_in_set('al.location_id', $ad_locations) . '
-				ORDER BY (' . $this->sql_random() . ' * a.ad_priority)
+						OR a.ad_end_date > " . time() . ")
+					$sql_where_views
+					$sql_where_clicks
+					AND " . $this->db->sql_in_set('al.location_id', $ad_locations) . '
+				ORDER BY (' . $this->sql_random() . ' * a.ad_priority) DESC
```
VSEphpbb: Are you sure this is DESC...or ASC (ASC is the default for ORDER BY)

VSEphpbb: Yeah just confirmed this should be ASC (but that's default so was OK without)

senky: First of all, `test_get_ads_priority` doesn't work properly then. It should fail now... Anyway. Are you sure ASC is the correct one? `array_filter` returns first ad for every location, throws all the other away. ASC will cause it to return the ad with the lowest priority (well, priority * random). DESC is supposed to do the opposite. Just discussion. Can you explain why ASC is the correct one?

senky:
!unset missing tests

VSEphpbb:
Well before we filtered it was ASC which worked with the old code which used the last item in the array. But if your filter flips that then this should be DESC.

senky:
@VSEphpbb yes it does. See? That's why I don't like filtering it in PHP. It's difficult to read. But maybe it's just my way of writing the code. Anyway, looks like it works now.

What does not is SQLite again! Agrrr.

VSEphpbb:
Something like https://stackoverflow.com/questions/19601948/must-appear-in-the-group-by-clause-or-be-used-in-an-aggregate-function

ad/manager.php
VSEphpbb: OK then DESC is OK

language/en/acp.php
```diff
@@ -20,35 +20,43 @@
...
+	'ACP_ADS_ADD'				=> 'Add new advertisement',
+	'ACP_ADS_EDIT'				=> 'Edit advertisement',
+	'AD_PREVIEW'				=> 'Preview this advertisement',
+	'CONFIGURE_AD'				=> 'Configure advertisement',
 
 	'AD_NAME_REQUIRED'			=> 'Name is required.',
 	'AD_NAME_TOO_LONG'			=> 'Name length is limited to %d characters.',
 	'AD_END_DATE_INVALID'		=> 'The end date is invalid or has already expired.',
 	'AD_PRIORITY_INVALID'		=> 'The priority is invalid. Please set a number between 1 and 10.',
+	'AD_VIEWS_LIMIT_INVALID'	=> 'The views limit is invalid. Please, set a non-negative number.',
```
VSEphpbb: no commas needed

migrations/v10x/m9_views_clicks.php
```diff
@@ -0,0 +1,99 @@
...
+		return array(
+			'add_columns' => array(
+				$this->table_prefix . 'ads' => array(
+					'ad_views'			=> array('UINT', 0),
+					'ad_clicks'			=> array('UINT', 0),
+					'ad_views_limit'	=> array('UINT', 0),
+					'ad_clicks_limit'	=> array('UINT', 0),
+				),
+			),
+			'add_index'	=> array(
```
VSEphpbb: Please do not add these indexes. They are unnecessary and they are also the reason why SQLite is failing to pass tests and function.

See https://github.com/phpbb-extensions/ad-management/pull/46

senky:
> Something like https://stackoverflow.com/questions/19601948/must-appear-in-the-group-by-clause-or-be-used-in-an-aggregate-function

@VSEphpbb looks really good! The only problem is, that we need to aggregate `location_id`, what is not a number, so it's really difficult to find an aggregate function. I hoped there exists something like `ID()`, which will simply return the same string (used often in functional programming). But there is nothing like this available. Any suggestion?

styles/all/template/event/memberlist_view_content_append.html
VSEphpbb: @senky Just squash this template's 2 set calls down like the other templates and this PR should be good.

PR #48, opened 2017-07-16
Ad owner

senky:
Admin can mark any registered user as advertisement owner. The user will be able to see advertisement statistics (views and clicks). This feature must be developed after “Views and clicks”.


controller/admin_controller.php
```diff
@@ -247,6 +271,12 @@ public function action_edit()
 			}
 			else if (empty($this->errors))
 			{
+				// Delete u_phpbb_ads_owner permission if old owner owns no more ads
+				if (count($this->manager->get_all_ads($ad['ad_owner'])) == 1)
```
VSEphpbb: I don't like that the get_all_ads query is being run here. This could probably be better handled instead by checking if the user has this permission already using something like `acl_get` or `al_get_list`

Actually you may even be able to run the acl_delete code below without checking at all, and not have any issues. You should check that out too.

controller/admin_controller.php
```diff
@@ -326,6 +356,12 @@ public function action_delete()
 				$this->manager->delete_ad_locations($ad_id);
 				$success = $this->manager->delete_ad($ad_id);
 
+				// Delete u_phpbb_ads_owner permission if old owner owns no more ads
+				if (count($this->manager->get_all_ads($ad_data['ad_owner'])) == 1)
```
VSEphpbb: Same comments as above

language/en/acp.php
```diff
@@ -45,6 +45,8 @@
 		1 => 'Click to disable',
 	),
 	'AD_EXPIRED_EXPLAIN'		=> 'This advertisement has expired and has been disabled.',
+	'AD_OWNER'					=> 'Owner',
+	'AD_OWNER_EXPLAIN'			=> 'Select an user who will be able to see statistics of this advertisement.',
```
VSEphpbb: "The ad owner is a user who can view the advertisement views and clicks statistics in the User Control Panel."

Also because that is the case, the setting for ad owner should be grouped in the ACP with the views and clicks settings.

language/en/acp.php
```diff
@@ -57,6 +59,7 @@
 	'AD_PRIORITY_INVALID'		=> 'The priority is invalid. Please set a number between 1 and 10.',
 	'AD_VIEWS_LIMIT_INVALID'	=> 'The views limit is invalid. Please set a non-negative number.',
 	'AD_CLICKS_LIMIT_INVALID'	=> 'The clicks limit is invalid. Please set a non-negative number.',
+	'AD_OWNER_INVALID'			=> 'The ad owner is invalid. Please select user using Find a member link below the field.',
```
VSEphpbb: "The ad owner is invalid. Please select a user using the Find a member link below the field."

language/en/permissions_phpbb_ads.php
```diff
@@ -0,0 +1,23 @@
...
+	exit;
+}
+
+if (empty($lang) || !is_array($lang))
+{
+	$lang = array();
+}
+
+$lang = array_merge($lang, array(
+	'ACL_U_PHPBB_ADS_OWNER' => 'Can see advertisement UCP module',
```
VSEphpbb: Most user permissions refer to the UCP so that is inferred, so should say:
"Can view advertisement statistics"

migrations/v10x/m10_ad_owner.php
```diff
@@ -0,0 +1,90 @@
...
+	{
+		return $this->db_tools->sql_column_exists($this->table_prefix . 'ads', 'ad_owner');
+	}
+
+	/**
+	 * {@inheritDoc}
+	 */
+	static public function depends_on()
+	{
+		return array('\phpbb\ads\migrations\v10x\m1_initial_schema');
```
VSEphpbb: this also depends on clicks and views

migrations/v10x/m10_ad_owner.php
```diff
@@ -0,0 +1,90 @@
...
+			),
+		);
+	}
+
+	/**
+	 * Add the UCP module and new permission
+	 *
+	 * @return array Array of data update instructions
+	 */
+	public function update_data()
```
VSEphpbb: This should be in its own migration (generally best to separate schema from data migrations)

styles/prosilver/template/ucp_ads_stats.html
```diff
@@ -0,0 +1,36 @@
+{% include 'ucp_header.html' %}
+
+<h2>{{ lang('UCP_PHPBB_ADS_STATS') }}</h2>
+
+<div class="panel">
+	<div class="inner">
+		{% if loops.ads %}
+		<ul class="topiclist">
+			<li class="header">
+				<dl class="row-item">
```
VSEphpbb: Styling is a bit off. Please take another look at UCP lists. For example, noticing that cplist and responsive classes are missing. Styling should probably just follow the Manage remember me login keys page as an example.

adm/style/manage_ads.html
```diff
@@ -48,6 +48,11 @@
 				<dd><input type="number" id="ad_priority" name="ad_priority" value="{{ AD_PRIORITY }}" min="1" max="10" placeholder="{{ constant('\\phpbb\\ads\\controller\\admin_controller::DEFAULT_PRIORITY') }}" /></dd>
 			</dl>
 			<dl>
+				<dt><label for="ad_owner">{{ lang('AD_OWNER') ~ lang('COLON') }}</label><br /><span>{{ lang('AD_OWNER_EXPLAIN') }}</span></dt>
+				<dd><input class="text medium" type="text" id="ad_owner" name="ad_owner" value="{{ AD_OWNER }}" /></dd>
+				<dd><strong><a href="{U_FIND_USERNAME}" onclick="find_username(this.href); return false;">{L_FIND_USERNAME}</a></strong></dd>
+			</dl>
```
VSEphpbb: Update: the above may not be working just because the URL was assigned wrong parameters which I mentioned in an earlier comment

ad/manager.php
```diff
@@ -294,6 +299,42 @@ public function load_groups()
...
+
+	/**
+	 * Get user_id from username
+	 *
+	 * @param	int		$user_id	User ID
+	 * @return	mixed	Username matching user_id
+	 */
+	public function get_owner_username_from_id($user_id)
+	{
+		$sql = 'SELECT username
```
senky: It should be `username` - the same thing that user inputs in the form. It is then cleaned using  `utf8_clean_string` in `get_owner_id_from_username`.

What I would like to do is replace those 2 methods with `user_get_id_name()` function from `functions_user.php`, but I am not sure whether global functions are what we want...

controller/admin_controller.php
senky: User might own more than one ad. So imagine a scenario when user1 owns 3 ads, admin edits one of his ads and assigns it to another user. Now, user1 owns 2 ads, but can't see them, because his permission was deleted. That's why I compare to one ad - if this is his only ad, only then we can delete the permission.

controller/admin_controller.php
```diff
@@ -255,21 +285,20 @@ public function action_edit()
...
 		}
 		else
 		{
-			// Load ad data
-			$data = $this->manager->get_ad($ad_id);
-			if (empty($data))
-			{
-				$this->error('ACP_AD_DOES_NOT_EXIST');
-			}
+			$data = $ad;
```
senky: This is not a mere duplication. See the code above - firstly, current ad data are loaded to `$ad`. Then, new ad data are assigned to `$data`. If there are no errors, we need old data to delete permission for old owner and new data to feed SQL query.

If you have a better idea how to treat this, I would be glad to rewrite the code.

migrations/v10x/m10_ad_owner.php
```diff
@@ -0,0 +1,90 @@
...
+	 *
+	 * @return array Array of table schema
+	 * @access public
+	 */
+	public function update_schema()
+	{
+		return array(
+			'add_columns' => array(
+				$this->table_prefix . 'ads' => array(
+					'ad_owner' => array('UINT', 0),
```
senky: I was always thinking about only one owner. Do you think multiple owners can be potentially so popular, that it is worth rewriting the code, adding new table, etc?

ad/manager.php
VSEphpbb: Yes you can use functions from functions_user...it should already be loaded and available to use and is preferred if these types of functions already exist in some form.

controller/admin_controller.php
VSEphpbb: OK well I see 2 options here:
1. Create a simple `is_ad_owner($user_id, $ad_id = 0)` function that simply returns the number of ads a user is owner of (and can also be used to check a specific ad if id is provided, just in case that is ever useful).

2. I prefer this. Lets NOT use any permissions any more for ad owners. Yes this means in the UCP all users will see the Advertisements tab ( @Derky ) but this should be intended. A user with no ads would just see the message `You have no advertisements.` in that tab. We do this for so many things already (like notifications). But think about how useful that tab could also be!? We could use that UCP Ads tab for the banner upload form. And we could even add a way for users to submit or request advertising from the board admin through that tab. I think that would be much more useful and less complicated by handling permissions.

controller/admin_controller.php
VSEphpbb: I see, then just add a comment to it explaining what it is doing.

controller/admin_controller.php
Derky: @VSEphpbb number 2 is a good idea indeed!

migrations/v10x/m10_ad_owner.php
VSEphpbb: I think it would be better to have multiple available. Sort of like in Titania, users added on separate lines (and stored as a JSON array. @Derky ?

Derky: @VSEphpbb @senky Let's keep it to one ad owner now. That's how Jakub described it and if you store it in JSON you can't easily filter the table on a user_id (for the list in the UCP) or join the users table in the ads list in the ACP.

VSEphpbb:
Ok

controller/admin_controller.php
```diff
@@ -562,64 +601,78 @@ protected function prepare_end_date($end_date)
...
 	{
 		$this->log->add('admin', $this->user->data['user_id'], $this->user->ip, 'ACP_PHPBB_ADS_' . $action . '_LOG', time(), array($ad_name));
 	}
+
+	/**
+	 * Get auth_admin class
+	 *
+	 * @return \auth_admin
+	 */
+	protected function get_auth_admin()
```
VSEphpbb: This can go away

language/en/ucp.php
```diff
@@ -0,0 +1,29 @@
...
+if (empty($lang) || !is_array($lang))
+{
+	$lang = array();
+}
+
+$lang = array_merge($lang, array(
+	'AD_NAME'	=> 'Name',
+	'AD_VIEWS'	=> 'Views',
+	'AD_CLICKS'	=> 'Clicks',
+	'NO_ADS'	=> 'No advertisements to display. When you become an owner of one, it will immediately display its statistics here.',
```
VSEphpbb: I think we can just say "You have no advertisements." It could also be strong I think.

language/en/ucp.php
```diff
@@ -0,0 +1,29 @@
...
+
+	'AD_VIEWS_DISABLED'		=> 'Advertisement views are not counted on this board, so they are not display in the table.',
```
VSEphpbb: For these lines, I think we should say, "Advertisement view counts are disabled on this board." Same for "click counts" However I'm not sure we need these lines at all. Why are we telling users about a statistic that the board admin has disabled?? We don't want to create a scenario where users start bugging their board admins to turn on clicks or views.

styles/prosilver/template/ucp_ads_stats.html
```diff
@@ -0,0 +1,31 @@
...
+			</thead>
+			<tbody>
+			{% for ad in loops.ads %}
+			<tr class="{% if loop.index0 is even %}bg2{% else %}bg1{% endif %}">
+				<td>{{ ad.NAME }}</td>
+				{% if S_VIEWS_ENABLED %}<td class="center">{{ ad.VIEWS }}</td>{% endif %}
+				{% if S_CLICKS_ENABLED %}<td class="center">{{ ad.CLICKS }}</td>{% endif %}
+			</tr>
+			{% else %}
+				<tr><td colspan="3" class="bg1" style="text-align: center">{{ lang('NO_ADS') }}</td></tr>
```
VSEphpbb: colspan is variable based on if views and/or clicks are enabled. Try:
```
colspan="{{ 1 + S_VIEWS_ENABLED + S_CLICKS_ENABLED }}"
```

language/en/ucp.php
senky: I would like to provide user with better explanation why there are no ads, what he can do about it, etc. This is what phpBB lacks overall, I think. A little bit of UX.

VSEphpbb: But this is not a good explanation. There may not be any stats, for example. If anything further is warranted it should  be "Users who own advertisements displayed on this board can view their statistics here."

controller/admin_controller.php
```diff
@@ -43,33 +43,40 @@ class admin_controller
...
+	protected $root_path;
+
 	/** @var string php_ext */
 	protected $php_ext;
 
 	/** @var string ext_path */
 	protected $ext_path;
 
+	/** @var \auth_admin Auth admin */
+	protected $auth_admin;
```
VSEphpbb: This can go away

controller/admin_controller.php
```diff
@@ -514,38 +530,55 @@ protected function get_form_data($form_name)
...
+
+			'AD_NAME'         => $data['ad_name'],
+			'AD_NOTE'         => $data['ad_note'],
+			'AD_CODE'         => $data['ad_code'],
+			'AD_ENABLED'      => $data['ad_enabled'],
+			'AD_END_DATE'     => $this->prepare_end_date($data['ad_end_date']),
+			'AD_PRIORITY'     => $data['ad_priority'],
+			'AD_VIEWS_LIMIT'  => $data['ad_views_limit'],
+			'AD_CLICKS_LIMIT' => $data['ad_clicks_limit'],
+			'AD_OWNER'        => isset($ad_owner_name[(int) $data['ad_owner'][0]]) ? $ad_owner_name[(int) $data['ad_owner'][0]] : $data['ad_owner'][0],
```
VSEphpbb: When editing an ad, if there is no owner, a `0` shows up in the Ad Owner field, but it should be `''`

controller/admin_controller.php
```diff
@@ -545,9 +528,17 @@ protected function get_form_data($form_name)
 		}
 
 		// Validate ad owner
-		if ($data['ad_owner'] === false)
+		if (!empty($data['ad_owner']))
+		{
+			user_get_id_name($ad_owner_id, $data['ad_owner']);
+			if (!empty($data['ad_owner']) && (!count($ad_owner_id) || ($data['ad_owner'] = $ad_owner_id[0]) === false))
```
Derky: I prefer to move the assignment of `$data['ad_owner'] = $ad_owner_id[0]` into an else-clause so it's clear that the variable is being changed and not just checked.

The comment above `// Validate ad owner` should ideally also say you're replacing the username with the user_id in the `$data['ad_owner']` var.

controller/admin_controller.php
```diff
@@ -601,6 +590,26 @@ protected function prepare_end_date($end_date)
 	}
 
 	/**
+	 * Prepare ad owner for display
+	 *
+	 * @param	mixed	$ad_owner	End date.
```
Derky: Incorrect var description (and the function description could use some better clarification what it's doing too)

controller/admin_controller.php
```diff
@@ -601,6 +590,26 @@ protected function prepare_end_date($end_date)
...
+	 * @return	string	End date prepared for display.
```
Derky: Incorrect var description

controller/admin_controller.php
```diff
@@ -601,6 +590,26 @@ protected function prepare_end_date($end_date)
...
+	protected function prepare_ad_owner($ad_owner)
+	{
+		if ($ad_owner)
+		{
+			user_get_id_name($ad_owner, $ad_owner_name);
+			if (empty($ad_owner_name))
+			{
+				return $ad_owner[0];
+			}
+			return $ad_owner_name[(int) $ad_owner[0]];
```
Derky: If the user was deleted, this will throw an undefined index I think?

language/en/ucp.php
VSEphpbb: Also in addition to above I think this should be styled with `strong`

controller/admin_controller.php
VSEphpbb: I just deleted a user assigned to an ad and what happens is the old user's ID (still stored in our ad's DB) appears in the ad owner's field when you go back to edit the ad the user was an owner of.

It might pay to also add an event to user delete where we remove the user's ID from our AD too.

VSEphpbb: Also if you're gonna use an else clause as @Derky suggested, since I never like multiple returns, it could also be nice to assign a variable the value to be returned and just have one return at the end of the function.

senky: @VSEphpbb we are already 3 days behind schedule, so later: #51

PR #49, opened 2017-07-17
#44: Make 5 a default value in priority field

### LINKED ISSUE #44: Make 5 a default value in priority field
Really annoying when creating ads that this field MUST be filled even if you don't intend to use it.

senky:
Resolves #44


controller/admin_controller.php
```diff
@@ -214,6 +214,7 @@ public function action_add()
 		else
 		{
 			$this->assign_locations();
+			$this->template->assign_var('AD_PRIORITY', self::DEFAULT_PRIORITY);
```
VSEphpbb: Actually don't do this. Just edit the template, swap the placeholder to the actual field value.

senky: That won't work when you edit an ad.

VSEphpbb:
This only matters for new ads. Editing an ad will already have a value. You can just use a conditional in the template. If it's empty use the default.

PR #52, opened 2017-07-21
Banner upload

senky:
Admin can upload image banner directly from ACP. The extension will automatically generate HTML code. Resize feature can also be included.


VSEphpbb:
The big issue with this is it's relying on a class that was introduced in phpBB 3.2 and did not exist in phpBB 3.1, so this is no longer 3.1 compatible.

DavidIQ:
I would say, given we've EOM'd 3.1, it's safe to only support 3.2 at this point for official extensions.

VSEphpbb:
I agree with @DavidIQ

senky:
@VSEphpbb glad to hear that. 😛 Will change composer.json as well then.

VSEphpbb:
@senky there should be a separate Pr for that, and it should also make other changes for phpBB 3.2, such as using the language class instead of user->lang

senky:
!unset WIP

senky:
!unset missing tests

adm/style/event/acp_overall_footer_after.html
```diff
@@ -36,6 +36,42 @@
...
+					url: window.location.href,
+					data: formData,
+					processData: false,
+					async: true,
+					cache: false,
+					contentType: false,
+					success: function(data) {
+						if (data.success) {
+							$('#ad_code').val(function(i, text) {
+								return text + "\n\n" + data.text;
```
VSEphpbb: We should check if field has text already or not before including the new lines, otherwise you get two unwanted empty new lines. This also applies to the PHP execution as well. e.g.

```
return (text.length ? text + "\n\n" : '') + data.text;
```

adm/style/manage_ads.html
```diff
@@ -44,6 +44,13 @@
 				<dd><textarea id="ad_code" name="ad_code" rows="25" cols="60" style="width: 95%;">{{ AD_CODE }}</textarea></dd>
 			</dl>
 			<dl>
+				<dt><label for="banner">{{ lang('BANNER') ~ lang('COLON') }}</label><br /><span>{{ lang('BANNER_EXPLAIN') }}</span></dt>
+				<dd>
+					<input type="file" accept="image/*" id="banner" name="banner" />
+					<button id="upload_banner" name="upload_banner">{{ lang('BANNER_UPLOAD') }}</button>
```
VSEphpbb: This needs proper ACP button styling classes added to it.

adm/style/manage_ads.html
```diff
@@ -44,6 +44,13 @@
 				<dd><textarea id="ad_code" name="ad_code" rows="25" cols="60" style="width: 95%;">{{ AD_CODE }}</textarea></dd>
 			</dl>
 			<dl>
+				<dt><label for="banner">{{ lang('BANNER') ~ lang('COLON') }}</label><br /><span>{{ lang('BANNER_EXPLAIN') }}</span></dt>
+				<dd>
+					<input type="file" accept="image/*" id="banner" name="banner" />
```
VSEphpbb: Add `class="inputbox autowidth"`

language/en/acp.php
```diff
@@ -52,6 +52,7 @@
 	'ACP_ADS_EDIT'				=> 'Edit advertisement',
 	'AD_PREVIEW'				=> 'Preview this advertisement',
 	'CONFIGURE_AD'				=> 'Configure advertisement',
+	'BANNER_UPLOAD'				=> 'Upload banner',
```
VSEphpbb: No translations yet for BANNER and BANNER_EXPLAIN

controller/admin_controller.php
```diff
@@ -454,11 +470,65 @@ protected function ad_enable($enable)
...
+		$this->files_upload->reset_vars();
+		$this->files_upload->set_allowed_extensions(array('gif', 'jpg', 'jpeg', 'png'));
+
+		// Upload file
+		$file = $this->files_upload->handle_upload('files.types.form', 'banner');
+		$file->clean_filename('unique_ext');
+		$file->move_file('images/phpbb_ads');
+
+		// Problem with uploading
+		if (sizeof($file->error))
```
VSEphpbb: Use count instead of sizeof

controller/admin_controller.php
```diff
@@ -454,11 +470,65 @@ protected function ad_enable($enable)
...
+	protected function process_banner_upload($ad_code)
+	{
+		// Set file restrictions
+		$this->files_upload->reset_vars();
+		$this->files_upload->set_allowed_extensions(array('gif', 'jpg', 'jpeg', 'png'));
+
+		// Upload file
+		$file = $this->files_upload->handle_upload('files.types.form', 'banner');
+		$file->clean_filename('unique_ext');
+		$file->move_file('images/phpbb_ads');
```
VSEphpbb: If `images/banner_ads` dir does not exist, we get the result of success, although nothing was uploaded.

`$file->move_file()` returns true/flase if file is uploaded so we need to use that to verify success or not. We also need to create the `images/phpbb_ads` directory if it does not exist yet. This `move_file()` method will quit and return false if the destination does not exist.

This class also has an `is_image()` method we should use as a check of the file's mimetype for added security.

controller/admin_controller.php
```diff
@@ -454,11 +470,65 @@ protected function ad_enable($enable)
...
 	/**
-	 * Get admin form data.
+	 * Upload image and return updated ad code or <img> of new banner when using ajax.
 	 *
-	 * @param    string $form_name The form name.
-	 * @return    array    Form data
+	 * @param	 string	 $ad_code	 Current ad code
+	 * @return	 mixed	 JsonResponse when request is ajax or updated ad code otherwise.
 	 */
+	protected function process_banner_upload($ad_code)
```
VSEphpbb: Is there a better place for this method to live? Or is that going to be done when you refactor this class?

VSEphpbb: Also just a thought... Have you tried or thought about instead of using this technique...using phpBB's plupload implementation and just letting users drag-drop images into the ad code text box to upload them...sort of the way it works when posting attachments?

controller/admin_controller.php
```diff
@@ -454,11 +470,65 @@ protected function ad_enable($enable)
...
+		if (sizeof($file->error))
+		{
+			$file->remove();
+			if ($this->request->is_ajax())
+			{
+				$json_response = new \phpbb\json_response;
+				$json_response->send(array(
+					'success'	=> false,
+					'title'		=> $this->user->lang('INFORMATION'),
+					'text'		=> implode(',', $file->error),
```
VSEphpbb: Shouldn't the glue in all the implodes be `', '` with a space after the comma?

controller/admin_controller.php
```diff
@@ -454,11 +470,65 @@ protected function ad_enable($enable)
...
+				));
+			}
+			else
+			{
+				$this->errors[] = implode(',', $file->error);
+			}
+		}
+		else
+		{
+			$banner_html = '<img src="' . generate_board_url() . '/images/phpbb_ads/' . $file->get('realname') . '" />';
```
VSEphpbb: Actually I think it would be better to check if `$ad_code` is empty or not here, and append the two empty newlines if it's not empty here. Then you can just return `$banner_html` below and in the javascript, you can also remove the empty new lines and just return `data.text`. In other words, let's fully form the banner html code here, just once.

controller/admin_controller.php
```diff
@@ -454,11 +470,65 @@ protected function ad_enable($enable)
 	}
 
 	/**
-	 * Get admin form data.
+	 * Upload image and return updated ad code or <img> of new banner when using ajax.
 	 *
-	 * @param    string $form_name The form name.
-	 * @return    array    Form data
+	 * @param	 string	 $ad_code	 Current ad code
+	 * @return	 mixed	 JsonResponse when request is ajax or updated ad code otherwise.
```
VSEphpbb: Actually it's not `JsonResponse` but `\phpbb\json_response`

controller/admin_controller.php
VSEphpbb: Actually instead of $banner_html you can just update the existing $ad_code and then have the one last return $ad_code at the end of the method to return the updated $ad_code.

adm/style/manage_ads.html
senky: Yeah, I tried, but they are bound to `input` or `a`. Don't know who wrote the selectors, but they block this button being styles.

VSEphpbb: You can make this an input type button then?

controller/admin_controller.php
senky: It is troublesome when you only append to existing $ad_code. The reason is, $ad_code is already html-encoded. So I would have to decode it on either php or js side before repopulating textarea. Let me see how I solved it with new commit and decide if it's ok or you still insist on appending to $ad_code.

controller/admin_controller.php
```diff
@@ -484,10 +484,17 @@ protected function process_banner_upload($ad_code)
 		// Upload file
 		$file = $this->files_upload->handle_upload('files.types.form', 'banner');
 		$file->clean_filename('unique_ext');
-		$file->move_file('images/phpbb_ads');
+		if (!$file->move_file('images/phpbb_ads'))
+		{
+			$file->set_error($this->user->lang('NO_UPLOAD_DIRECTORY'));
+		}
+		if (!$file->is_image())
```
VSEphpbb: We dont need this. It's already been uploaded at this point, so no point in checking it. It seems this has only been used in phpBB before removing an image file.

language/en/acp.php
```diff
@@ -47,6 +47,8 @@
 	'AD_EXPIRED_EXPLAIN'		=> 'This advertisement has expired and has been disabled.',
 	'AD_OWNER'					=> 'Owner',
 	'AD_OWNER_EXPLAIN'			=> 'The ad owner is a user who can view the advertisement views and clicks statistics in the User Control Panel.',
+	'BANNER'					=> 'Upload a banner',
+	'BANNER_EXPLAIN'			=> 'Select an image and click the button next to it to generate HTML IMG tag with the image for your advertisement code.',
```
VSEphpbb: `Upload a banner image`

I'd rather say something like:
`Upload an image in JPG, GIF or PNG format. Image will be stored in phpBB's <samp>images</samp> directory and an HTML IMG tag for the image will automatically be inserted into the ad code field.`

migrations/v10x/m12_banner_dir.php
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	/**
+	 * Create images/phpbb_ads directory
+	 *
+	 * @return void
+	 */
+	public function create_banner_dir()
+	{
+		mkdir($this->phpbb_root_path . 'images/phpbb_ads', 0777, true);
```
VSEphpbb: I do not like this. Migrations are meant for database changes.

For a filesystem change, we should do that in PHP. So our upload banner should have a means to check if dir exists, and create on if it does not instead. Besides, it's always good in our upload method to make sure the destination exists first anyways, before we try moving files around.

controller/admin_controller.php
VSEphpbb: Ah, OK

controller/admin_controller.php
```diff
@@ -484,6 +484,15 @@ protected function process_banner_upload($ad_code)
...
 		$file->clean_filename('unique_ext');
+		$dir = $this->root_path . 'images/phpbb_ads';
+		if (!file_exists($dir))
+		{
+			mkdir($this->root_path . 'images/phpbb_ads', 0777, true);
+			if (!file_exists($dir))
+			{
+				$file->set_error($this->user->lang('NO_UPLOAD_DIRECTORY'));
+			}
+		}
```
VSEphpbb:
```php
// First lets create phpbb_ads directory if needed
if (!file_exists($dir))
{
   try
   {
      $filesystem->mkdir($dir);
   }
   catch (\phpbb\filesystem\exception\filesystem_exception $e)
   {
      $file->set_error($e->get_message());
   }
}
```

controller/admin_controller.php
```diff
@@ -484,9 +489,25 @@ protected function process_banner_upload($ad_code)
...
 		$file = $this->files_upload->handle_upload('files.types.form', 'banner');
 		$file->clean_filename('unique_ext');
+
+		// First lets create phpbb_ads directory if needed
+		$dir = $this->root_path . 'images/phpbb_ads';
+		if (!$this->filesystem->exists($dir))
+		{
+			try
+			{
+				$this->filesystem->mkdir($this->root_path . 'images/phpbb_ads');
```
VSEphpbb: just use `$dir`

controller/admin_controller.php
```diff
@@ -484,9 +489,25 @@ protected function process_banner_upload($ad_code)
...
+			{
+				$file->set_error($this->user->lang($e->getMessage()));
+			}
+		}
+
+		// Move file to proper location
 		if (!$file->move_file('images/phpbb_ads'))
 		{
-			$file->set_error($this->user->lang('NO_UPLOAD_DIRECTORY'));
+			$file->set_error($this->user->lang('CANNOT_CREATE_DIRECTORY'));
```
VSEphpbb: CANNOT_CREATE_DIRECTORY should have to do with failing to create phpbb_ads. This error message should have to do with failing to copy the file to the phpbb_ads directory.

language/en/acp.php
```diff
@@ -64,8 +64,7 @@
 	'AD_CLICKS_LIMIT_INVALID'	=> 'The clicks limit is invalid. Please set a non-negative number.',
 	'AD_OWNER_INVALID'			=> 'The ad owner is invalid. Please select a user using the Find a member link.',
 	'NO_FILE_SELECTED'			=> 'No file selected.',
-	'NO_UPLOAD_DIRECTORY'		=> 'File was not uploaded. Check upload directory: images/phpbb_ads.',
-	'NOT_IMAGE'					=> 'File is not an image.',
+	'CANNOT_CREATE_DIRECTORY'	=> 'File was not uploaded. Check upload directory: <samp>images/phpbb_ads</samp>.',
```
VSEphpbb: OK, so based on the translation, `NO_UPLOAD_DIRECTORY` was OK here. 
`CANNOT_CREATE_DIRECTORY` should be: The phpbb_ads directory could not be created. Please make sure the phpbb/images directory is writable."

controller/admin_controller.php
senky: I tried, but it doesn't work. It apparently expects only relative path.

VSEphpbb: Using $dir works for me.

controller/admin_controller.php
```diff
@@ -505,7 +505,7 @@ protected function process_banner_upload($ad_code)
 		}
 
 		// Move file to proper location
-		if (!$file->move_file('images/phpbb_ads'))
+		if (!$file->move_file($dir))
```
VSEphpbb: This needs to be reverted

PR #54, opened 2017-07-24
#50: Refactor admin controller

### LINKED ISSUE #50: Refactor admin controller
Admin controller is becoming dangerously big. I am thinking of ways to refactor it. One solution might be creating a separate classes for every form field. Each such class would request a variable, validate inputted data and prepare it for display in the template. This way, we could potentially keep admin controller untouched when adding new fields. What do you think?

senky:
Resolves #50

### DIFF
--- a/banner/banner.php
+++ b/banner/banner.php
@@ -45,6 +45,8 @@ class banner
 
 	/**
 	 * Create storage directory for banners uploaded by Ads Management
+	 *
+	 * @throws \phpbb\filesystem\exception\filesystem_exception
 	 */
 	public function create_storage_dir()
 	{
@@ -56,6 +58,9 @@ class banner
 
 	/**
 	 * Handle banner upload
+	 *
+	 * @throws	\phpbb\exception\runtime_exception
+	 * @return	string	Filename
 	 */
 	public function upload()
 	{
--- a/banner/banner.php
+++ b/banner/banner.php
@@ -1,0 +1,91 @@
+<?php
+/**
+ *
+ * Advertisement management. An extension for the phpBB Forum Software package.
+ *
+ * @copyright (c) 2017 phpBB Limited <https://www.phpbb.com>
+ * @license GNU General Public License, version 2 (GPL-2.0)
+ *
+ */
+
+namespace phpbb\ads\banner;
+
+class banner
+{
+	/** @var \phpbb\files\upload */
+	protected $files_upload;
+
+	/** @var \phpbb\filesystem\filesystem_interface */
+	protected $filesystem;
+
+	/** @var string */
+	protected $root_path;
+
+	/** @var \phpbb\files\filespec */
+	protected $file;
+
+	/**
+	 * Constructor
+	 *
+	 * @param \phpbb\files\upload						$files_upload	Files upload object
+	 * @param \phpbb\filesystem\filesystem_interface	$filesystem		Filesystem object
+	 * @param string									$root_path		Root path
+	 */
+	public function __construct(\phpbb\files\upload $files_upload, \phpbb\filesystem\filesystem_interface $filesystem, $root_path)
+	{
+		$this->files_upload = $files_upload;
+		$this->filesystem = $filesystem;
+		$this->root_path = $root_path;
+	}
+
+	public function set_file($file)
+	{
+		$this->file = $file;
+	}
+
+	/**
+	 * Create storage directory for banners uploaded by Ads Management
+	 */
+	public function create_storage_dir()
+	{
+		if (!$this->filesystem->exists($this->root_path . 'images/phpbb_ads'))
+		{
+			$this->filesystem->mkdir($this->root_path . 'images/phpbb_ads');
+		}
+	}
+
+	/**
+	 * Handle banner upload
+	 */
+	public function upload()
+	{
+		// Set file restrictions
+		$this->files_upload->reset_vars();
+		$this->files_upload->set_allowed_extensions(array('gif', 'jpg', 'jpeg', 'png'));
+
+		// Upload file
+		$this->set_file($this->files_upload->handle_upload('files.types.form', 'banner'));
+		$this->file->clean_filename('unique_ext');
+
+		// Move file to proper location
+		if (!$this->file->move_file('images/phpbb_ads'))
+		{
+			$this->file->set_error('FILE_MOVE_UNSUCCESSFUL');
+		}
+
+		if (count($this->file->error))
+		{
+			throw new \phpbb\exception\runtime_exception($this->file->error[0]);
+		}
+
+		return $this->file->get('realname');
+	}
+
+	/**
+	 * Remove file from the filesystem
+	 */
+	public function remove()
+	{
+		$this->file->remove();
+	}
+}
--- a/controller/admin_controller.php
+++ b/controller/admin_controller.php
@@ -15,9 +15,7 @@ namespace phpbb\ads\controller;
 */
 class admin_controller
 {
-	const MAX_NAME_LENGTH = 255;
 	const DATE_FORMAT = 'Y-m-d';
-	const DEFAULT_PRIORITY = 5;
 
 	/** @var \phpbb\template\template */
 	protected $template;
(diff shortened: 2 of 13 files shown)


controller/admin_controller.php
```diff
@@ -106,11 +76,7 @@ public function __construct(\phpbb\template\template $template, \phpbb\user $use
 	public function mode_manage()
 	{
 		$this->setup();
-
-		if (!function_exists('user_get_id_name'))
-		{
-			include($this->root_path . 'includes/functions_user.' . $this->php_ext);
-		}
+		$this->input->setup();
```
VSEphpbb: I don't like that there are two of these setups. I would just add this include of the functions_user to the setup() that's in the admin_controller, so there's only one setup.

controller/admin_controller.php
```diff
@@ -15,9 +15,7 @@
 */
 class admin_controller
 {
-	const MAX_NAME_LENGTH = 255;
 	const DATE_FORMAT = 'Y-m-d';
```
VSEphpbb: I think this constant can also be moved to the the input class with the others.

controller/admin_controller.php
```diff
@@ -213,32 +171,33 @@ public function action_add()
...
 			{
 				$this->ad_preview($data['ad_code']);
 			}
 			else if ($upload_banner)
 			{
-				$data['ad_code'] = $this->process_banner_upload($data['ad_code']);
+				$data['ad_code'] = $this->input->process_banner_upload($data['ad_code']);
 			}
-			else if (empty($this->errors))
+			else if (empty($this->input->get_errors()))
```
VSEphpbb: This is not allowed and causing the tests to fail. Consider perhaps a `input->has_errors()` method that returns a boolean.

controller/admin_controller.php
```diff
@@ -266,17 +225,17 @@ public function action_edit()
...
```
VSEphpbb: Same issue, has_errors() instead.

controller/admin_helper.php
```diff
@@ -0,0 +1,163 @@
...
+	protected function prepare_end_date($end_date)
+	{
+		if (empty($end_date))
+		{
+			return '';
+		}
+
+		if (is_numeric($end_date))
+		{
+			return $this->user->format_date($end_date, self::DATE_FORMAT);
```
VSEphpbb: DATE_FORMAT is not in this class. It should be in the input class with the other constants.

controller/admin_input.php
```diff
@@ -0,0 +1,271 @@
...
+	{
+		$this->user = $user;
+		$this->request = $request;
+		$this->files_upload = $files_upload;
+		$this->filesystem = $filesystem;
+		$this->root_path = $root_path;
+		$this->php_ext = $php_ext;
+	}
+
+	public function setup()
```
VSEphpbb: Docblocks are needed for all methods in this class.

VSEphpbb: Oh and this method should be removed, it's code moved into the setup() in the controller

controller/admin_input.php
```diff
@@ -0,0 +1,271 @@
...
+	{
+		if (!function_exists('user_get_id_name'))
+		{
+			include($this->root_path . 'includes/functions_user.' . $this->php_ext);
+		}
+	}
+
+	public function get_errors()
```
VSEphpbb: Also have a has_errors class, something simple like:

```php
public function has_errors()
{
    return count($this->errors);
}
```

controller/admin_input.php
```diff
@@ -0,0 +1,271 @@
...
+		return (int) $this->user->get_timestamp_from_format(controller::DATE_FORMAT, $end_date);
+	}
+
+	protected function owner_to_id($ad_owner)
+	{
+		if (empty($ad_owner))
+		{
+			return 0;
+		}
+		else
```
VSEphpbb: Else is not needed really.

controller/admin_input.php
```diff
@@ -0,0 +1,271 @@
...
+		{
+			user_get_id_name($ad_owner_id, $ad_owner);
+			return $ad_owner_id[0];
+		}
+	}
+}
```
VSEphpbb: Missing new line at EOF

controller/admin_input.php
```diff
@@ -0,0 +1,271 @@
...
+		return $data;
+	}
+
+	/**
+	 * Upload image and return updated ad code or <img> of new banner when using ajax.
+	 *
+	 * @param	 string	 $ad_code	 Current ad code
+	 * @return	 mixed	 \phpbb\json_response when request is ajax or updated ad code otherwise.
+	 */
+	public function process_banner_upload($ad_code)
```
VSEphpbb: Can we just call this "banner_upload" ?

controller/admin_controller.php
senky: `DATE_FORMAT` is really a difficult case. It's present in all 3 classes, but most occurrences are in `admin_controller`, that's why it kept it there. Any idea?

senky:
@Derky @VSEphpbb any idea how to decrease cyclomatic complexity of `banner_upload` to reach at least B score?

controller/admin_controller.php
VSEphpbb: It doesn't matter where it is since its used in multiple places, which is why it's best to kep them all together. You could store them all in ext.php too, if you wanted. That's what I do when I have constants used in multiple places.

controller/admin_input.php
```diff
@@ -49,29 +46,24 @@ class admin_input
...
 		$this->user = $user;
 		$this->request = $request;
 		$this->files_upload = $files_upload;
 		$this->filesystem = $filesystem;
 		$this->root_path = $root_path;
-		$this->php_ext = $php_ext;
 	}
 
-	public function setup()
+	public function get_errors()
```
VSEphpbb: Missing docblocks

controller/admin_input.php
```diff
@@ -112,7 +104,10 @@ public function get_form_data($form_name)
 
 		// Replace end date and owner with IDs that will be stored in the DB
 		$data['ad_end_date'] = $this->end_date_to_timestamp($data['ad_end_date']);
-		$data['ad_owner'] = $this->owner_to_id($data['ad_owner']);
+		if (!in_array($this->user->lang('AD_OWNER_INVALID'), $this->errors))
```
VSEphpbb: I don't like this, checking for a translated string.

I see 2 things needing to be done here.

1. send only language keys to the errors array. Then call the lang() method in one place, right before we display them to the user (in assign_errors). It's OK if some errors are already translated (such as from exceptions thrown by phpBB) because they will just pass through the lang() method unharmed.

2. Then here just check if array key exists for `AD_OWNER_INVALID`.

This will also simplufy all your error assignments since no more need to call lang() on each one.

controller/admin_input.php
```diff
@@ -134,16 +129,13 @@ public function process_banner_upload($ad_code)
 		$file->clean_filename('unique_ext');
 
 		// First lets create phpbb_ads directory if needed
-		if (!$this->filesystem->exists($this->root_path . 'images/phpbb_ads'))
+		try
```
VSEphpbb: If you want ot extract creating a storage directory into it's own method, then this whole try/catch block should be part of that method, so anything that goes wrong with it, is part of its own contained method.

That way you can eliminate the comment "First lets create phpbb_ads directory if needed" as the method call itself will make it obvious what ishappening here.

controller/admin_input.php
```diff
@@ -152,41 +144,23 @@ public function process_banner_upload($ad_code)
 			$file->set_error($this->user->lang('FILE_MOVE_UNSUCCESSFUL'));
 		}
 
+		$error = count($file->error);
```
VSEphpbb: I'd drop this error down below banner_html

VSEphpbb:
@senky reduce conditionals. Moving out the whole try/catch block into the create_storage_dir will help, as I mentioned in review.

VSEphpbb:
maybe make banner uploading its own class with simple methods... create storage dir, upload image, throw exceptions on errors, then that would vastly shrink and simplify the banner_upload code inside your input class. It could become something like:

```php
public function banner_upload($ad_code)
{
	try
	{
		$this->banner->create_storage_dir();
		$this->banner->upload();
	}
	catch (\phpbb\exceptions\runtime_exception $e)
	{
		$file->remove();
		
		if ($this->request->is_ajax())
		{
			$this->send_ajax_response(false, $e->get_message());
		}
	}

	$banner_html = '<img src="' . generate_board_url() . '/images/phpbb_ads/' . $file->get('realname') . '" />';

	if ($this->request->is_ajax())
	{
		$this->send_ajax_response(true, $banner_html);
	}

	return ($ad_code ? $ad_code . "\n\n" : '') . $banner_html;
}
```

banner/banner.php
```diff
@@ -0,0 +1,91 @@
...
+		$this->root_path = $root_path;
+	}
+
+	public function set_file($file)
+	{
+		$this->file = $file;
+	}
+
+	/**
+	 * Create storage directory for banners uploaded by Ads Management
```
VSEphpbb: Should have `@throws \phpbb\filesystem\exception\filesystem_exception`

banner/banner.php
```diff
@@ -0,0 +1,91 @@
...
+	{
+		if (!$this->filesystem->exists($this->root_path . 'images/phpbb_ads'))
+		{
+			$this->filesystem->mkdir($this->root_path . 'images/phpbb_ads');
+		}
+	}
+
+	/**
+	 * Handle banner upload
+	 */
```
VSEphpbb: Should have `@throws \phpbb\exception\runtime_exception`

controller/admin_input.php
```diff
@@ -0,0 +1,238 @@
...
+				$this->send_ajax_response(false, $this->user->lang($e->getMessage()));
+			}
+
+			$this->errors[] = $this->user->lang($e->getMessage());
+		}
+
+		return $ad_code;
+	}
+
+	protected function validate_ad_name($ad_name)
```
VSEphpbb: Still hoping for docblocks on these methods

controller/admin_controller.php
```diff
@@ -10,15 +10,13 @@
 
 namespace phpbb\ads\controller;
 
+use \phpbb\ads\controller\admin_input as input;
```
VSEphpbb: The leading slash is causing tests fails and should be removed

controller/admin_helper.php
```diff
@@ -0,0 +1,167 @@
...
+ * Advertisement management. An extension for the phpBB Forum Software package.
+ *
+ * @copyright (c) 2017 phpBB Limited <https://www.phpbb.com>
+ * @license GNU General Public License, version 2 (GPL-2.0)
+ *
+ */
+
+namespace phpbb\ads\controller;
+
+use \phpbb\ads\controller\admin_input as input;
```
VSEphpbb: Remove leading slash

controller/admin_helper.php
```diff
@@ -0,0 +1,167 @@
...
+		// Returns false when no errors occur trying to find the user
+		if (false === user_get_id_name($ad_owner, $ad_owner_name))
+		{
+			if (empty($ad_owner_name))
+			{
+				return $ad_owner[0];
+			}
+			return $ad_owner_name[(int) $ad_owner[0]];
+		}
+		return '';
```
VSEphpbb: This function has a little weirdness. user_get_id_name() kind of expects it's params to be arrays... not int and string.

Maybe we should do this:
```php
$user_id = array($ad_owner);
$user_name = array();

// Returns false when no errors occur trying to find the user
if (false === user_get_id_name($user_id, $user_name))
{
	if (empty($user_name))
	{
		return $user_id[0];
	}
	return $user_name[(int) $user_id[0]];
}
return '';
```

senky: It's only incomplete docblocks that's causing phpStorm to complain. Function easily handles either single input or an array: https://github.com/phpbb/phpbb/blob/3.2.x/phpBB/includes/functions_user.php#L45-L50

But alright, let's make it easier to read.

PR #56, opened 2017-07-27
Refactor for phpBB 3.2. Drop support of phpBB 3.1.

senky:
Track here: https://trello.com/c/emM3GbN5/42-refactor-for-phpbb-32-drop-support-of-phpbb-31


adm/style/manage_ads.html
```diff
@@ -1,5 +1,8 @@
 {% include 'overall_header.html' %}
 
+{% INCLUDECSS '@phpbb_ads/phpbbads.css' %}
```
VSEphpbb: This is only needed for previewing ads, so it could go under `{% if S_ADD_AD or S_EDIT_AD %}`

adm/style/manage_ads.html
```diff
@@ -1,5 +1,8 @@
...
+{% INCLUDECSS '@phpbb_ads/datetimepicker/jquery.datetimepicker.min.css' %}
```
VSEphpbb: This is only used with the datepicker so it could go with `{% INCLUDEJS '@phpbb_ads/datetimepicker/jquery.datetimepicker.full.min.js' %}`

config/location.yml
```diff
@@ -16,68 +16,78 @@ services:
         class: phpbb\ads\location\type\above_footer
         arguments:
             - '@user'
+            - '@language'
```
VSEphpbb: You can also, if you want, create a service for the location base class: `phpbb.ads.location.type.base`.
Then every child location class here can replace
```
arguments:
	- '@user'
	- '@language'
```
with
```
parent: phpbb.ads.location.type.base
```
The nice thing about that is if we ever change the base dependencies, you only have to edit the base service.

styles/all/template/event/overall_header_content_before.html
```diff
@@ -1,3 +1,14 @@
+{% INCLUDECSS '@phpbb_ads/phpbbads.css' %}
+{% if S_PHPBB_ADS_ENABLE_CLICKS %}
+	{% INCLUDEJS '@phpbb_ads/js/clicks.js' %}
+{% endif %}
```
VSEphpbb: These two conditions are the same and can be combined into one.

config/location.yml
```diff
@@ -12,72 +12,69 @@ services:
         tags:
             - { name: service_collection, tag: phpbb.ads.location.type }
 
-    phpbb.ads.location.type.above_footer:
-        class: phpbb\ads\location\type\above_footer
+    phpbb.ads.location.type.base:
```
VSEphpbb: Since this base is not actually a tagged location, lets move it up above the `# ----- Template locations -----` line

composer.json
```diff
@@ -24,21 +24,21 @@
...
-			"filename": "version_check"
-		}
+        "version-check": {
+            "host": "www.phpbb.com",
+            "directory": "/customise/db/extension/ads",
+            "filename": "version_check"
+        }
     }
-}
+}
```
VSEphpbb: Missine new line at end. Also please use tabs in composer.json. (phpBB uses tabs in theirs).

adm/style/event/acp_overall_footer_after.html
```diff
@@ -1,7 +1,8 @@
 {% if S_PHPBB_ADS %}
-	{% if not definition.INCLUDED_DATETIMEPICKERJS %}
+	{% if not definition.INCLUDED_DATETIMEPICKER %}
```
VSEphpbb: We can move all the code in this template event out of here and eliminate it. We only need all this code in our `manage_ads.html` so lets just put it there.

PR #61, opened 2017-07-31
Remove request from controller

VSEphpbb:
@senky what do you think of this? I moved all requests to the input class and away from the controller.


senky:
I thought about it as well, just didn't see copying 3 request methods over to `input` as a good idea.

PR #67, opened 2017-08-06
#65: Improve precise views

### LINKED ISSUE #65: Improve precise views
Obfuscate the URL so that crawlers won't find it when regex-ing source code.

senky:
Resolves #65

### DIFF
--- a/event/main_listener.php
+++ b/event/main_listener.php
@@ -119,7 +119,9 @@ class main_listener implements EventSubscriberInterface
 			{
 				$this->template->assign_vars(array(
 					'S_INCREMENT_VIEWS'		=> true,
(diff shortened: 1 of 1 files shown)


VSEphpbb:
How does this work?

senky:
`(![]+[])[+!+[]]` returns `'a'`, so final string for the JS is `/app.php/' + (![]+[])[+!+[]] + 'dsview/1-2-3` which is then translated to `/app.php/adsview/1-2-3` before calling `$.get`.

It's not very clear on PHP side. But I didn't find any way to force controller helper return obfuscated URL already. I need to postprocess it with `str_replace`.

PR #69, opened 2017-08-06
Ad snippet analysis

senky:
Analysis of entered snippet will be presented to the admin, such as “untrusted connection” when board runs on HTTPS but advertisement snippet will try to load HTTP JS file, “blocking javascript” when `<script>` tag without `async` attribute will be present, etc.

### DIFF
--- a/adm/style/manage_ads.html
+++ b/adm/style/manage_ads.html
@@ -42,7 +42,27 @@
 			</dl>
 			<dl>
 				<dt><label for="ad_code">{{ lang('AD_CODE') ~ lang('COLON') }}</label><br /><span>{{ lang('AD_CODE_EXPLAIN') }}</span></dt>
-				<dd><textarea id="ad_code" name="ad_code" rows="20" cols="60" style="width: 95%;">{{ AD_CODE }}</textarea></dd>
+				<dd>
+					<textarea id="ad_code" name="ad_code" rows="20" cols="60" style="width: 95%;">{{ AD_CODE }}</textarea>
+					<input type="submit" class="button2" id="analyse_ad_code" name="analyse_ad_code" value="{{ lang('ANALYSE_AD_CODE') }}">
+					<div class="analyser-results">
+						{% if loops.analyser_results_notice or loops.analyser_results_warning %}
+							{% if loops.analyser_results_notice %}
+								{% for notice in loops.analyser_results_notice %}
+									<p class="warningbox">{{ notice.MESSAGE }}</p>
+								{% endfor %}
+							{% endif %}
+
+							{% if loops.analyser_results_warning %}
+								{% for warning in loops.analyser_results_warning %}
+									<p class="errorbox">{{ warning.MESSAGE }}</p>
+								{% endfor %}
+							{% endif %}
+						{% elseif CODE_ANALYSED %}
+							<p class="successbox">{{ lang('EVERYTHING_OK') }}</p>
+						{% endif %}
+					</div>
+				</dd>
 			</dl>
 		</fieldset>
 		<fieldset>
--- a/analyser/manager.php
+++ b/analyser/manager.php
@@ -1,0 +1,79 @@
+<?php
+/**
+ *
+ * Advertisement management. An extension for the phpBB Forum Software package.
+ *
+ * @copyright (c) 2017 phpBB Limited <https://www.phpbb.com>
+ * @license GNU General Public License, version 2 (GPL-2.0)
+ *
+ */
+
+namespace phpbb\ads\analyser;
+
+class manager
+{
+	/** @var array Ad code analysis tests */
+	protected $tests;
+
+	/** @var \phpbb\request\request */
+	protected $request;
+
+	/** @var \phpbb\template\template */
+	protected $template;
+
+	/** @var \phpbb\language\language */
+	protected $lang;
+
+	/**
+	 * Construct an ad code analysis manager object
+	 *
+	 * @param	array						$tests		Ad code analysis tests passed via the service container
+	 * @param	\phpbb\request\request		$request	Request object
+	 * @param	\phpbb\template\template	$template	Template object
+	 * @param	\phpbb\language\language	$lang		Language object
+	 */
+	public function __construct($tests, \phpbb\request\request $request, \phpbb\template\template $template, \phpbb\language\language $lang)
+	{
+		$this->tests = $tests;
+		$this->request = $request;
+		$this->template = $template;
+		$this->lang = $lang;
+	}
+
+	/**
+	 * Test the ad code for potential problems.
+	 *
+	 * @param	string	$ad_code	Advertisement code
+	 */
+	public function run($ad_code)
+	{
+		$results = array();
+		foreach ($this->tests as $test)
+		{
+			$result = $test->run($ad_code);
+			if ($result !== false)
+			{
+				$results[] = $result;
+			}
+		}
+
+		$this->assign_template_vars($results);
+	}
+
+	/**
+	 * Assign analyser results to template variables.
+	 *
+	 * @param	array	$results	Analyser results
+	 */
+	protected function assign_template_vars($results)
+	{
+		foreach ($results as $result)
+		{
+			$this->template->assign_block_vars('analyser_results_' . $result['severity'], array(
+				'MESSAGE'	=> $this->lang->lang($result['message']),
+			));
+		}
+
+		$this->template->assign_var('CODE_ANALYSED', true);
+	}
+}
--- a/analyser/test/alert.php
+++ b/analyser/test/alert.php
@@ -1,0 +1,35 @@
+<?php
+/**
+ *
+ * Advertisement management. An extension for the phpBB Forum Software package.
+ *
+ * @copyright (c) 2017 phpBB Limited <https://www.phpbb.com>
+ * @license GNU General Public License, version 2 (GPL-2.0)
+ *
+ */
+
+namespace phpbb\ads\analyser\test;
+
+class alert implements test_interface
+{
+	/**
+	 * {@inheritDoc}
+	 *
+	 * Javascript alert() test.
+	 * This test checks for the presence of alert() in an ad code.
+	 * There is no reason why ad would trigger alert, so it's
+	 * categorized as warning.
+	 */
+	public function run($ad_code)
+	{
+		if (preg_match('/alert\s*\(/U', $ad_code))
+		{
+			return array(
+				'severity'	=> 'warning',
+				'message'	=> 'ALERT_USAGE',
+			);
+		}
+
+		return false;
+	}
+}
--- a/analyser/test/script_without_async.php
+++ b/analyser/test/script_without_async.php
@@ -1,0 +1,41 @@
+<?php
+/**
+ *
+ * Advertisement management. An extension for the phpBB Forum Software package.
+ *
+ * @copyright (c) 2017 phpBB Limited <https://www.phpbb.com>
+ * @license GNU General Public License, version 2 (GPL-2.0)
+ *
+ */
+
+namespace phpbb\ads\analyser\test;
+
+class script_without_async implements test_interface
+{
+	/**
+	 * {@inheritDoc}
+	 *
+	 * Synchronously loaded scripts test.
+	 * This test looks for scripts that aren't using `async` attribute
+	 * to load itself asynchronously. Such scripts slow down page rendering
+	 * time and should be made asynchronous.
+	 */
+	public function run($ad_code)
+	{
+		if (preg_match_all('/&lt;script(.*)src(.*)&gt;/U', $ad_code, $matches))
+		{
+			foreach ($matches[1] as $match)
+			{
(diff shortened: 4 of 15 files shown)


senky:
!set WIP

senky:
!unset WIP

adm/style/manage_ads.html
```diff
@@ -42,7 +42,34 @@
 			</dl>
 			<dl>
 				<dt><label for="ad_code">{{ lang('AD_CODE') ~ lang('COLON') }}</label><br /><span>{{ lang('AD_CODE_EXPLAIN') }}</span></dt>
-				<dd><textarea id="ad_code" name="ad_code" rows="20" cols="60" style="width: 95%;">{{ AD_CODE }}</textarea></dd>
+				<dd>
+					<textarea id="ad_code" name="ad_code" rows="20" cols="60" style="width: 95%;">{{ AD_CODE }}</textarea>
+					<p>{{ lang('AD_CODE_NOT_SURE') }} <input type="submit" class="button2" id="analyse_ad_code" name="analyse_ad_code" value="{{ lang('ANALYSE_AD_CODE') }}"></p>
```
VSEphpbb: I think all we need here is the button (dont need the NOT_SURE line)

adm/style/manage_ads.html
```diff
@@ -42,7 +42,34 @@
...
+							{% if loops.analyser_results_warning %}
+								<h4>{{ lang('WARNINGS') }}</h4>
+								<ul>
+									{% for warning in loops.analyser_results_warning %}
+										<li>{{ warning.MESSAGE }}</li>
+									{% endfor %}
+								</ul>
+							{% endif %}
+						{% else %}
+							{{ lang('EVERYTHING_OK') }}
```
VSEphpbb: We probably shouldn't say anything here. If the ad code has not been analyzed, we shouldn't presume to say the code is ok.

adm/style/manage_ads.html
```diff
@@ -42,7 +42,34 @@
...
+									<li>{{ notice.MESSAGE }}</li>
+								{% endfor %}
+								</ul>
+							{% endif %}
+
+							{% if loops.analyser_results_warning %}
+								<h4>{{ lang('WARNINGS') }}</h4>
+								<ul>
+									{% for warning in loops.analyser_results_warning %}
+										<li>{{ warning.MESSAGE }}</li>
```
VSEphpbb: There's a little inconsistency in the indentations between these two blocks of code here and above.

We should only say something if the test has been run and passed (maybe define this template var that everything is OK only if there are no test fails, after it has been run).

analyser/manager.php
```diff
@@ -0,0 +1,64 @@
...
+	 */
+	public function __construct($tests, \phpbb\request\request $request, \phpbb\template\template $template, \phpbb\language\language $lang)
+	{
+		$this->tests = $tests;
+		$this->request = $request;
+		$this->template = $template;
+		$this->lang = $lang;
+	}
+
+	public function test($ad_code)
```
VSEphpbb: docblock?

VSEphpbb: I would prefer calling this run. Also in the following classes, calling it run. It looks better when we have:
```
$test->run();
$analyser->run();
```

analyser/test/alert.php
```diff
@@ -0,0 +1,31 @@
...
+ * @license GNU General Public License, version 2 (GPL-2.0)
+ *
+ */
+
+namespace phpbb\ads\analyser\test;
+
+class alert implements test_interface
+{
+	/**
+	 * {@inheritDoc}
```
VSEphpbb: I'd like to see updated ineherited doc blocks here that explain the test. e.g.,

```
/**
 * {@inheritDoc}
 *
 * Javascript alert() test.
 * This test checks for the presence of alert() in an ad code because it would be 
 * very bad...very bad indeed.
 */
```

analyser/test/alert.php
```diff
@@ -0,0 +1,31 @@
...
+	 */
+	public function test($ad_code)
+	{
+		if (preg_match('/alert(\s)*\(/U', $ad_code))
```
VSEphpbb: Do you need the capturing groups? Can you just use:
```
/alert\s*\(/U
```

adm/style/manage_ads.html
```diff
@@ -42,7 +42,34 @@
 			</dl>
 			<dl>
 				<dt><label for="ad_code">{{ lang('AD_CODE') ~ lang('COLON') }}</label><br /><span>{{ lang('AD_CODE_EXPLAIN') }}</span></dt>
-				<dd><textarea id="ad_code" name="ad_code" rows="20" cols="60" style="width: 95%;">{{ AD_CODE }}</textarea></dd>
+				<dd>
+					<textarea id="ad_code" name="ad_code" rows="20" cols="60" style="width: 95%;">{{ AD_CODE }}</textarea>
+					<p>{{ lang('AD_CODE_NOT_SURE') }} <input type="submit" class="button2" id="analyse_ad_code" name="analyse_ad_code" value="{{ lang('ANALYSE_AD_CODE') }}"></p>
+					<div class="analyser-results">
+						{% if loops.analyser_results_notice or loops.analyser_results_warning %}
+							<h3>{{ lang('RESULTS') }}</h3>
```
VSEphpbb: For this I think `h2` looks nice

We should also do some better styling to the output below. Like make the  more sever warning results red and the less sever notices can probably be default text colors.

language/en/acp.php
```diff
@@ -29,6 +29,11 @@
 	'AD_NOTE_EXPLAIN'			=> 'Enter any notes for this advertisement. These notes are not shown anywhere except in the ACP and are optional.',
 	'AD_CODE'					=> 'Code',
 	'AD_CODE_EXPLAIN'			=> 'Enter the advertisement code here. All code must use HTML markup, BBCodes are not supported.',
+	'AD_CODE_NOT_SURE'			=> 'Not sure whether the code is OK? You can',
+	'ANALYSE_AD_CODE'			=> 'analyse advertisement code',
```
VSEphpbb: Capitalize `Analyse`

language/en/acp.php
```diff
@@ -29,6 +29,11 @@
...
+	'RESULTS'					=> 'Results',
+	'NOTICES'					=> 'Notices',
+	'EVERYTHING_OK'				=> 'The code seems to be OK.',
```
VSEphpbb: The code appears OK.

language/en/acp.php
```diff
@@ -80,6 +85,12 @@
 	'ACP_AD_DISABLE_SUCCESS'	=> 'Advertisement disabled successfully.',
 	'ACP_AD_DISABLE_ERRORED'	=> 'There was an error disabling the advertisement.',
 
+	// Analyser tests
+	'UNSECURE_CONNECTION'	=> '<strong>Mixed Content</strong><br />Your code will probably generate “Mixed Content” warning in the browser, because your board runs on HTTPS, but the ad code attempts to load a file from unsecure HTTP connection.',
```
VSEphpbb: <strong>Mixed Content</strong><br />Your board runs on a secure HTTPS connection, however the ad code is attempting to load content from an insecure HTTP connection. This can cause browsers to generate a "Mixed Content" warning to let users know that the page contains insecure resources.

language/en/acp.php
```diff
@@ -80,6 +85,12 @@
...
+	'SCRIPT_WITHOUT_ASYNC'	=> '<strong>script without async attribute</strong><br />Your code will load script synchronously, which affects page load performance. Use <samp>async</samp> attribute to speed up the page load.',
```
VSEphpbb: <strong>Non-asynchronous javascript</strong><br />This ad code loads JavaScript code in a non-asynchronous way. This means it will block any other Javascript from loading until it has completed loading, which can affect page load performance. Use of the <samp>async</samp> attribute can speed up the page load.

language/en/acp.php
```diff
@@ -80,6 +85,12 @@
...
+	'ALERT_USAGE'			=> '<strong>Usage of <samp>alert()</samp></strong><br />Your code uses <samp>alert()</samp> function which annoys users and even blocks them using your board before they close it. Some browsers will even block page load and display warning to the user.',
```
VSEphpbb: <strong>Usage of <samp>alert()</samp></strong><br />Your code uses the <samp>alert()</samp> function which is not a good practice and can distract users. Some browsers may also block page load and display additional warnings to the user.

language/en/acp.php
```diff
@@ -80,6 +85,12 @@
...
+	'LOCATION_CHANGE'		=> '<strong>Redirection</strong><br />Your code will redirect user to another (probably malicious) page.',
```
VSEphpbb: <strong>Redirection</strong><br />Your code appears it can redirect user to another page or site. Redirects can sometimes send users to unintended, often malicious, destinations. Please verify the integrity of your ad code's redirection destination.

analyser/test/script_without_async.php
```diff
@@ -0,0 +1,37 @@
...
+namespace phpbb\ads\analyser\test;
+
+class script_without_async implements test_interface
+{
+	/**
+	 * {@inheritDoc}
+	 */
+	public function test($ad_code)
+	{
+		if (preg_match_all('/&lt;script(.*)&gt;/U', $ad_code, $matches))
```
VSEphpbb: This test is failing because this first regex is matching the closing script tag as well, and since the closing tag has no async...false positive.

senky: Are you sure? It can't match closing tag, there is a slash after `&lt;` in that case...

VSEphpbb:
yep, positive. I had breakpoints and it ran through the condition twice…once for the opening tag and again for the closing tag.

senky:
@VSEphpbb can you post your ad code here?

VSEphpbb:
@senky I see what it is.. there's a script within a script in Google Adsense

```
<div class="stat-block" style="width:100%">
	<h3 style="margin-top:10px;">Advertisements</h3>
	<div  style="text-align: center; padding: 12px 0">
		<script async src="//pagead2.googlesyndication.com/pagead/js/adsbygoogle.js"></script>
		<!-- Blog Leaderboard -->
		<ins class="adsbygoogle"
		     style="display:inline-block;width:728px;height:90px"
		     data-ad-client="x
		     data-ad-slot="x"></ins>
		<script>
		(adsbygoogle = window.adsbygoogle || []).push({});
		</script>
	</div>
</div>
```

VSEphpbb:
@senky you should update your regex to only check for async in script tags that contain a `src` attribute I think

analyser/manager.php
```diff
@@ -0,0 +1,81 @@
...
+	 */
+	public function run($ad_code)
+	{
+		$result = array(
+			'notice'	=> array(),
+			'warning'	=> array(),
+		);
+		foreach ($this->tests as $test)
+		{
+			$result = array_merge_recursive($result, $test->run($ad_code));
```
VSEphpbb: Note that this triggers an inspection warning that merging arrays in a loop causes high CPU usage and is a resources greedy construction. Perhaps there is an alternative approach?

https://github.com/kalessil/phpinspectionsea/blob/master/docs/performance.md#slow-array-function-used-in-loop

adm/style/manage_ads.html
```diff
@@ -42,7 +42,30 @@
...
+							{% endif %}
+
+							{% if loops.analyser_results_warning %}
+								<h4>{{ lang('WARNINGS') }}</h4>
+									{% for warning in loops.analyser_results_warning %}
+										<p class="errorbox">{{ warning.MESSAGE }}</p>
+									{% endfor %}
+							{% endif %}
+						{% elseif CODE_ANALYSED %}
+							{{ lang('EVERYTHING_OK') }}
```
VSEphpbb: We might as well wrap this in a `<p class="successbox""></p>` so it looks like the other messages and makes users feel extra good about themselves.

adm/style/manage_ads.html
```diff
@@ -42,7 +42,30 @@
...
 				<dt><label for="ad_code">{{ lang('AD_CODE') ~ lang('COLON') }}</label><br /><span>{{ lang('AD_CODE_EXPLAIN') }}</span></dt>
-				<dd><textarea id="ad_code" name="ad_code" rows="20" cols="60" style="width: 95%;">{{ AD_CODE }}</textarea></dd>
+				<dd>
+					<textarea id="ad_code" name="ad_code" rows="20" cols="60" style="width: 95%;">{{ AD_CODE }}</textarea>
+					<input type="submit" class="button2" id="analyse_ad_code" name="analyse_ad_code" value="{{ lang('ANALYSE_AD_CODE') }}">
+					<div class="analyser-results">
+						{% if loops.analyser_results_notice or loops.analyser_results_warning %}
+							<h2 class="centered-text">{{ lang('RESULTS') }}</h2>
+							{% if loops.analyser_results_notice %}
+								<h4>{{ lang('NOTICES') }}</h4>
```
VSEphpbb: I doubt we need the Results, Warnings and Notices headings anymore. The newly stylized message boxes should be enough, and IMO look a little cleaner without all the headings now.

PR #73, opened 2017-08-09
Merge clicks and views controller to one

senky:


VSEphpbb:
This still isn't actually work. A functional test would discover this.
Basically getting the error:
Missing value for argument #2: data in class phpbb\ads\controller\increment_controller:handle
The ad_id parameter is being passed off to the handle() method as the $mode value. And no 2nd arg being sent to $data.

VSEphpbb:
Update: even though a functional test here would not be able to check success because of the ajax, it could test that the correct unauthorized access page is shown, which in this case is not happening because the route error is displayed instead.

config/routing.yml
```diff
@@ -1,11 +1,13 @@
...
     requirements:
+        mode: 'clicks'
         ad_id: '\d+'
 
 phpbb_ads_view:
     path: /adsview/{ad_ids}
-    defaults: { _controller: phpbb.ads.view.controller:increment_views }
+    defaults: { _controller: phpbb.ads.increment.controller:handle, mode: 'views' }
     requirements:
+        mode: 'views'
```
rxu: ```mode``` isn't defined within the ```path``` here and above, so will it actually be checked for this requirement then?

config/routing.yml
```diff
@@ -1,11 +1,13 @@
 phpbb_ads_click:
     path: /adsclick/{ad_id}
-    defaults: { _controller: phpbb.ads.click.controller:increment_clicks }
+    defaults: { _controller: phpbb.ads.increment.controller:handle, mode: 'clicks' }
     requirements:
+        mode: 'clicks'
         ad_id: '\d+'
```
VSEphpbb: all `ad_id` and `ad_ids` need to be changed to `data` to work.

Also, this still needs a functional test to verify it finally is working.

config/routing.yml
VSEphpbb: mode is defined in the route definition, so it can't be submitted via the url

VSEphpbb: @rxu you're right. the mode requirements are not needed.

VSEphpbb: also we don't need the quotes around clicks and views

config/routing.yml
VSEphpbb: Also we do not need quotes around clicks or views or the requirements

VSEphpbb: Simple tests like this should verify the route is working
```
$crawler = self::request('GET', 'app.php/adsclick/1', [], false);
$this->assertContainsLang('NOT_AUTHORISED', $crawler->text());
```

config/routing.yml
rxu: 😛

tests/functional/increment_views_test.php
```diff
@@ -0,0 +1,23 @@
...
+namespace phpbb\ads\tests\functional;
+
+/**
+ * @group functional
+ */
+class increment_views_test extends functional_base
+{
+	public function test_click_without_ajax()
+	{
+		$crawler = self::request('GET', 'app.php/adsview/1', [], false);
```
VSEphpbb: Please just combine these 2 files into 1 file with the 2 tests

PR #74, opened 2017-08-09
ACP Ad Management Improvements

senky:
https://trello.com/c/NaLWqyQJ/23-acp-ad-management-improvements

### DIFF
--- a/adm/style/manage_ads.html
+++ b/adm/style/manage_ads.html
@@ -150,18 +150,18 @@
 				},
 				{
 					'heading': lang('EXPIRED_ADS'),
-					'loop': loops.force_disabled,
+					'loop': loops.expired,
 					'allow_enable': false
 				}
 			] %}
-				{% if list['loop'] %}
-					<td class="row3" colspan="7"><strong>{{ list['heading'] }}</strong></td>
-					{% for ad in list['loop'] %}
-						<tr>
+				{% if list.loop %}
+					<td class="row3" colspan="7"><strong>{{ list.heading }}</strong></td>
+					{% for ad in list.loop %}
+						<tr{% if ad.S_EXPIRED %} title="{{ lang('AD_EXPIRED_EXPLAIN') }}"{% endif %}>
 							<td><strong>{{ ad.NAME }}</strong></td>
 							<td>{{ ad.PRIORITY }}</td>
 							<td>
-								{% if ad.S_END_DATE_EXPIRED %}<strong title="{{ lang('AD_EXPIRED_EXPLAIN') }}" class="error">{% endif %}
+								{% if ad.S_END_DATE_EXPIRED %}<strong class="error">{% endif %}
 									{{ ad.END_DATE }}
 								{% if ad.S_END_DATE_EXPIRED %}</strong>{% endif %}
 							</td>
@@ -189,13 +189,12 @@
 							<td><a href="{{ ad.U_EDIT }}">{{ ICON_EDIT }}</a> <a href="{{ ad.U_DELETE }}" data-ajax="row_delete">{{ ICON_DELETE }}</a></td>
 						</tr>
 					{% endfor %}
+				{% elseif loop.index0 %}
+					<tr>
+						<td colspan="7" style="text-align: center;">{{ lang('ACP_ADS_EMPTY') }}</td>
+					</tr>
 				{% endif %}
 			{% endfor %}
-			{% if S_NO_ADS %}
-			<tr>
-				<td colspan="7" style="text-align: center;">{{ lang('ACP_ADS_EMPTY') }}</td>
-			</tr>
-			{% endif %}
 		</tbody>
 	</table>
 
--- a/adm/style/manage_ads.html
+++ b/adm/style/manage_ads.html
@@ -133,9 +133,10 @@
 		<thead>
 			<tr>
 				<th>{{ lang('AD_NAME') }}</th>
-				<th style="width: 15%;">{{ lang('AD_END_DATE') }}</th>
-				{% if S_VIEWS_ENABLED %}<th style="width: 10%;">{{ lang('AD_VIEWS') }}</th>{% endif %}
-				{% if S_CLICKS_ENABLED %}<th style="width: 10%;">{{ lang('AD_CLICKS') }}</th>{% endif %}
+				<th style="width: 7%">{{ lang('AD_PRIORITY') }}</th>
+				<th style="width: 10%;">{{ lang('AD_END_DATE') }}</th>
+				{% if S_VIEWS_ENABLED %}<th style="width: 7%;">{{ lang('AD_VIEWS') }}</th>{% endif %}
+				{% if S_CLICKS_ENABLED %}<th style="width: 7%;">{{ lang('AD_CLICKS') }}</th>{% endif %}
 				<th style="width: 10%;">{{ lang('ENABLED') }}</th>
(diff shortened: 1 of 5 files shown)


adm/style/manage_ads.html
```diff
@@ -133,9 +133,11 @@
 		<thead>
 			<tr>
 				<th>{{ lang('AD_NAME') }}</th>
-				<th style="width: 15%;">{{ lang('AD_END_DATE') }}</th>
```
VSEphpbb: With these many columns, I would redo these percentages. The Ad Name and Ad Locations should each be 20%... Leave all the rest empty so they auto size to fit.

adm/style/manage_ads.html
```diff
@@ -145,6 +147,14 @@
 				<tr>
 					<td><strong>{{ ad.NAME }}</strong></td>
 					<td>
+						{% for location in ad.locations %}
+							{{ location.NAME }}<br />
+						{% else %}
+							{{ lang('NO_LOCATIONS') }}
```
VSEphpbb: This ahs no translation But I don't think it's necessary. It's more noticable when the field is just blank if there are no locations set.

controller/admin_controller.php
```diff
@@ -403,6 +405,16 @@ public function list_ads()
 				'U_EDIT'             => $this->u_action . '&amp;action=edit&amp;id=' . $row['ad_id'],
 				'U_DELETE'           => $this->u_action . '&amp;action=delete&amp;id=' . $row['ad_id'],
 			));
+
+			if (isset($ads_locations[$row['ad_id']]))
```
VSEphpbb: This might be a problem in some PHP7 cases (you'll get an error that ther'es an undefined index). Probably better to use `array_key_exists($row['ad_id'], $ads_locations)`

controller/admin_helper.php
```diff
@@ -149,6 +149,18 @@ public function prepare_end_date($end_date)
 		return (string) $end_date;
 	}
 
+	public function prepare_ads_locations($ads_locations)
+	{
+		$result = array();
+		$location_types = $this->location_manager->get_all_locations();
```
VSEphpbb: it's a bit confusing that there are two methods named `get_all_locations`... one in the `location_manager` and one in the `ad_manager`. I'd rename the new one in the ad manager to `get_all_ad_locations` since its not really getting all locations, it's getting all the ads's locations 😄

adm/style/manage_ads.html
senky: Personally, I think ad locations made this table too difficult to read. Maybe we can hide it somehow and display on hover or something...

VSEphpbb: Just remove it then

VSEphpbb: But we still should not be setting percentages on every column. Just set it on the most important column, which would be the ad name.

controller/admin_controller.php
```diff
@@ -379,25 +379,30 @@ public function action_delete()
 	 */
 	public function list_ads()
 	{
+		$no_ads = true;
```
VSEphpbb: Remove `$no_ads` please

controller/admin_controller.php
```diff
@@ -379,25 +379,30 @@ public function action_delete()
...
 		foreach ($this->manager->get_all_ads() as $row)
 		{
+			$no_ads = false;
 			$ad_enabled = (int) $row['ad_enabled'];
 			$ad_end_date = (int) $row['ad_end_date'];
-			$ad_expired = $ad_end_date > 0 && $ad_end_date < time();
+			$ad_end_date_expired = $ad_end_date > 0 && $ad_end_date < time();
```
VSEphpbb: We won't be needing this. Please just add this logic to `%ad_expired` so there is only 1 variable to represent expiration.

controller/admin_controller.php
```diff
@@ -379,25 +379,30 @@ public function action_delete()
...
 				'NAME'               => $row['ad_name'],
+				'PRIORITY'			 => $row['ad_priority'],
 				'END_DATE'           => $this->helper->prepare_end_date($ad_end_date),
 				'VIEWS'              => $row['ad_views'],
 				'CLICKS'             => $row['ad_clicks'],
 				'VIEWS_LIMIT'        => $row['ad_views_limit'],
 				'CLICKS_LIMIT'       => $row['ad_clicks_limit'],
-				'S_END_DATE_EXPIRED' => $ad_expired,
+				'S_EXPIRED' 		 => $ad_expired,
+				'S_END_DATE_EXPIRED' => $ad_end_date_expired,
```
VSEphpbb: This template var will not be needed. In the template, to highlight expired ad dates, we can use:
```twig
{% if ad.END_DATE < "now"|date("Y-m-d") %}
```
And perhaps to limit processing in the loop, we can set a TWIG variable before the loops with `"now"|date("Y-m-d")` and then use that.
```twig
{% set NOW = "now"|date("Y-m-d") %}
...
{% if ad.END_DATE < NOW %}
```

adm/style/manage_ads.html
```diff
@@ -132,48 +132,68 @@
...
+							{% if S_CLICKS_ENABLED %}
+								<td>
+									{% if ad.CLICKS_LIMIT and ad.CLICKS >= ad.CLICKS_LIMIT %}<strong class="error">{% endif %}
+										{{ ad.CLICKS }}{% if ad.CLICKS_LIMIT %} / {{ ad.CLICKS_LIMIT }}{% endif %}
+									{% if ad.CLICKS_LIMIT and ad.CLICKS >= ad.CLICKS_LIMIT %}</strong>{% endif %}
+								</td>
+							{% endif %}
+							<td>
+								{% if list['allow_enable'] %}
+									{% if not ad.S_END_DATE_EXPIRED %}<a href="{{ ad.U_ENABLE }}" title="{{ lang('AD_ENABLE_TITLE', ad.S_ENABLED) }}" data-ajax="toggle_enable">{% endif %}
```
VSEphpbb: This is no longer just about the date. This should be:
```twig
{% if list.allow_enable %}
	<a href="{{ ad.U_ENABLE }}" title="{{ lang('AD_ENABLE_TITLE', ad.S_ENABLED) }}" data-ajax="toggle_enable">
		{% if ad.S_ENABLED %}{{ lang('ENABLED') }}{% else %}{{ lang('DISABLED') }}{% endif %}
	</a>
{% else %}
	{% if ad.S_ENABLED %}{{ lang('ENABLED') }}{% else %}{{ lang('DISABLED') }}{% endif %}
{% endif %}
```

adm/style/manage_ads.html
```diff
@@ -180,10 +181,12 @@
 								</td>
 							{% endif %}
 							<td>
-								{% if list['allow_enable'] %}
-									{% if not ad.S_END_DATE_EXPIRED %}<a href="{{ ad.U_ENABLE }}" title="{{ lang('AD_ENABLE_TITLE', ad.S_ENABLED) }}" data-ajax="toggle_enable">{% endif %}
+								{% if list.allow_enable %}
+									{% if not ad.S_EXPIRED %}<a href="{{ ad.U_ENABLE }}" title="{{ lang('AD_ENABLE_TITLE', ad.S_ENABLED) }}" data-ajax="toggle_enable">{% endif %}
```
VSEphpbb: Why is this conditional still here? If an ad is expired, it won't be in the allow.enable group. So there are no need for conditionals around this anchor tags. So in other words, every thing in the `list.allow_enable` should should have this anchor tag.

PR #75, opened 2017-08-12
Improve Scrutinizer score

senky:


controller/admin_controller.php
```diff
@@ -433,6 +323,114 @@ protected function setup()
...
+				}
+
+				// Load ad template locations
+				$data['ad_locations'] = $this->manager->get_ad_locations($ad_id);
+			}
+		}
+
+		// Set output vars for display in the template
+		$this->template->assign_vars(array(
+			'S_ADD_AD'           => $action == 'add',
```
VSEphpbb: `"S_{$action}_AD" => $action` ??

controller/admin_controller.php
```diff
@@ -433,6 +323,114 @@ protected function setup()
...
+			'S_EDIT_AD'          => $action == 'edit',
+			'EDIT_ID'            => $ad_id,
```
VSEphpbb: Insetad of 'EDIT_IS' could just be 'AD_ID'

controller/admin_controller.php
```diff
@@ -433,6 +323,114 @@ protected function setup()
...
+			'U_BACK'             => $this->u_action,
+			'U_ACTION'           => "{$this->u_action}&amp;action=$action" . ($action == 'edit' ? '&amp;id=' . $ad_id : ''),
```
VSEphpbb: instead of `$action == 'edit'` should be `($ad_id ?  '&amp;id=' . $ad_id : '')`

controller/admin_controller.php
```diff
@@ -381,29 +307,17 @@ public function list_ads()
 	{
 		foreach ($this->manager->get_all_ads() as $row)
 		{
-			$ad_enabled = (int) $row['ad_enabled'];
+			$ad_enabled  = (int) $row['ad_enabled'];
 			$ad_end_date = (int) $row['ad_end_date'];
```
VSEphpbb: $ad_end_date is no longer needed for anything

controller/admin_controller.php
```diff
@@ -224,114 +203,80 @@ public function action_add()
 
 		// Set output vars for display in the template
 		$this->template->assign_vars(array(
-			'S_ADD_AD'           => true,
-			'U_BACK'             => $this->u_action,
-			'U_ACTION'           => "{$this->u_action}&amp;action=add",
-			'PICKER_DATE_FORMAT' => input::DATE_FORMAT,
-			'U_FIND_USERNAME'    => $this->helper->get_find_username_link(),
+			'S_ADD_AD'				=> true,
```
VSEphpbb: This template block can be extracted from both the add and edit functions and combined into one method like:
```php
/**
 * Set output vars for display in the template
 * 
 * @param string $action add|edit
 * @param int    $ad_id  The ad id (optional and only needed for edit action)
 */
protected function assign_page_vars($action, $ad_id = 0)
{
		'S_' . strtoupper($action) . '_AD' => true,
		'AD_ID' => $ad_id,
		'PICKER_DATE_FORMAT' => input::DATE_FORMAT,
		'U_BACK' => $this->u_action,
		'U_ACTION' => "{$this->u_action}&amp;action={$action}" . ($ad_id ? "&amp;id=$ad_id" : ''),
		'U_FIND_USERNAME' => $this->helper->get_find_username_link(),
	));
}
```
Note that I changed the template var `EDIT_ID` to `AD_ID`. That better represents what the value it holds is and will just be ignored in the add page and only used on the edit page.

From add method:
```php
$this->assign_page_vars('add');
```

From edit method:
```php
$this->assign_page_vars('edit', $ad_id);
```

event/main_listener.php
```diff
@@ -149,4 +147,17 @@ public function remove_ad_owner($event)
 	{
 		$this->manager->remove_ad_owner($event['user_ids']);
 	}
+
+	protected function views($ad_ids)
```
VSEphpbb: I would leave this public (so it can be unit tested better) and move it to be above or below clicks, so they are together and not separated by the remove_ad_owner function. Also, docblocks.

event/main_listener.php
```diff
@@ -115,20 +115,18 @@ public function setup_ads()
 				));
 			}
 
-			if ($this->config['phpbb_ads_enable_views'] && !$this->user->data['is_bot'] && count($ad_ids))
```
VSEphpbb: The first few lines of setup_ads() can be extracted too:
```php
/**
 * User can view ads if they are not in a group that has ads hidden
 *
 * @return bool true if the user is not in a group with ads hidden, false if they are
 */
protected function can_view_ads()
{
	$user_groups = $this->manager->load_memberships($this->user->data['user_id']);
	$hide_groups = json_decode($this->config_text->get('phpbb_ads_hide_groups'), true);

	return !array_intersect($user_groups, $hide_groups);
}
```

event/main_listener.php
```diff
@@ -115,20 +115,18 @@ public function setup_ads()
...
-					// Obfuscate URL to prevent crawlers increasing view counters.
-					// Uses http://www.jsfuck.com/ to make 'a' really complicated, yet executable.
-					'UA_PHPBB_ADS_VIEWS'	=> str_replace('adsview', "' + (![]+[])[+!+[]] + 'dsview", $this->controller_helper->route('phpbb_ads_view', array('data' => implode('-', $ad_ids)))),
-				));
-			}
+			$this->views($ad_ids);
 		}
+	}
 
+	public function adblocker()
```
VSEphpbb: docblocks :)

controller/admin_controller.php
senky: I did it that way before. But then I decided not to extract every single almost identical code into separate method, because then it became enormously robust with too much methods. So current state is basically a compromise between good code practices and readability. At least for me. Do you feel this can improve code quality?

VSEphpbb: In this case I think it's OK mainly because it's got duplication. It's always better to have to change something in only 1 place :)

VSEphpbb:
Please address the issues Scrutinizer found: https://scrutinizer-ci.com/g/phpbb-extensions/ad-management/inspections/2de83692-2281-47ad-92b9-18163404aa1c/issues/

senky:
@VSEphpbb I did except for the last one which greatly reduces readability of the method, so I kept it as it is.

controller/admin_controller.php
```diff
@@ -366,43 +311,43 @@ public function action_delete()
...
+	protected function list_ads()
 	{
 		foreach ($this->manager->get_all_ads() as $row)
 		{
-			$ad_enabled = (int) $row['ad_enabled'];
-			$ad_end_date = (int) $row['ad_end_date'];
-			$ad_expired = ($ad_end_date > 0 && $ad_end_date < time()) || ($row['ad_views_limit'] && $row['ad_views'] >= $row['ad_views_limit']) || ($row['ad_clicks_limit'] && $row['ad_clicks'] >= $row['ad_clicks_limit']);
+			$ad_enabled  = (int) $row['ad_enabled'];
+			$ad_expired  = $this->helper->is_expired($row);
+
```
VSEphpbb: The lines below in `if ($ad_expired && $ad_enabled)` are not covered by any tests.

VSEphpbb:
@senky has merge conflicts

controller/admin_controller.php
```diff
@@ -17,6 +17,9 @@
 */
 class admin_controller
 {
+	/** @var array|bool Form data */
+	protected $data = false;
```
VSEphpbb: does this have to be false? Why not array()? Type switching can sometimes be confusing. It seems it can only be set to false by the return of `get_ad` which we specifically check for. Otherwise, we throw data into so many other functions that all expect it to be an array that it may be safer to avoid initializing it as false, and just always consider it an array (because the one time it's false we trigger error).

controller/admin_controller.php
```diff
@@ -465,33 +410,101 @@ protected function ad_enable($enable)
 	}
 
 	/**
-	 * Prepare advertisement preview
+	 * Submit action "preview".
+	 * Prepare advertisement preview.
+	 *
+	 * @return	void
+	 */
+	protected function preview()
```
VSEphpbb: Lol, I feel like the methods below could go in a sub-class for `admin_actions`. Just because this class in general has so many methods and some of them seem to be the similar (in name), i.e.: `action_add` and then `submit_add`. ANyway I think it's possible some of the protected methods in this class can be moved into their own logical sub classes???

migrations/v10x/m2_acp_module.php
```diff
@@ -33,8 +33,9 @@ public function effectively_installed()
 	 */
 	public static function depends_on()
 	{
-		return array('\phpbb\ads\migrations\v10x\m1_initial_schema');
-	}
+		return array(
+			'\phpbb\ads\migrations\v10x\m1_initial_schema',
+		);	}
```
VSEphpbb: why is this formatted weirdly

migrations/v10x/m3_template_locations_schema.php
```diff
@@ -25,7 +25,9 @@ public function effectively_installed()
 	 */
 	public static function depends_on()
 	{
-		return array('\phpbb\ads\migrations\v10x\m1_initial_schema');
+		return array(
+			'\phpbb\ads\migrations\v10x\m1_initial_schema',
+		);
```
VSEphpbb: why was this changed? or any of the other single item arrays for migrations?

migrations/v10x/m9_views_clicks.php
```diff
@@ -25,7 +25,16 @@ public function effectively_installed()
...
+		return array(
+			'\phpbb\ads\migrations\v10x\m1_initial_schema',
+			'\phpbb\ads\migrations\v10x\m2_acp_module',
+			'\phpbb\ads\migrations\v10x\m3_template_locations_schema',
+			'\phpbb\ads\migrations\v10x\m4_indexes',
+			'\phpbb\ads\migrations\v10x\m5_end_date',
+			'\phpbb\ads\migrations\v10x\m6_hide_for_group',
+			'\phpbb\ads\migrations\v10x\m7_adblocker',
+			'\phpbb\ads\migrations\v10x\m8_priority',
+		);
```
VSEphpbb: Why?

controller/admin_controller.php
senky: It's good idea. Nice separation of concerns. Can we leave it for later? I will depart soon and won't have time for this for the next 3 weeks. But will do after that, some mid-September and later. I have created new issue for this: #82.

VSEphpbb: of course. was just a new idea.

migrations/v10x/m9_views_clicks.php
senky: To be honest: I have no idea. But now you can uninstall the extension without errors.

ad/manager.php
```diff
@@ -55,6 +55,11 @@ public function get_ad($ad_id)
 		$data = $this->db->sql_fetchrow($result);
 		$this->db->sql_freeresult($result);
 
+		if ($data === false)
```
VSEphpbb: just 1 line:

```
return $data !== false ? $data : array();
```

PR #77, opened 2017-08-13
Better lloking buttons in ACP

VSEphpbb:


styles/all/theme/phpbbads.css
```diff
@@ -5,18 +5,45 @@
...
 	width: 30px;
-	background-color: white;
+	cursor: pointer;
+}
+
+.phpbb-ads-button {
+	font-size: 13px;
+	vertical-align: middle;
+	background: #e3e3e3;
+	background: -moz-linear-gradient(top, #e3e3e3 0%, #f7f7f7 100%);
```
senky: Oh please, remove all of them. Keep only `background: linear-gradient(to bottom, #e3e3e3 0%, #f7f7f7 100%);`.  Chrome, Firefox, Opera and Edge already support official syntax and I don' feel like supporting old IEs...

PR #79, opened 2017-08-15
Address validation issues

senky:
- [ ] 1) Please add a screenshot to the CDB so users can see what the extensions does at a glance.

- [x] 2) From ad/manager.php:
```
			/* All other cases should use the default
			case 'mssql':
			case 'mssql_odbc':
			case 'mssqlnative':
			case 'mysql':
			case 'mysqli':*/
			default:
```
Remove commented code. If you want to make your point, just have it as a list.

- [x] 3) `phpbb.alert('{{ lang('INFORMATION') }}', '{{ lang('NO_FILE_SELECTED') }}');`
Language items need to be escaped in javascript.

- [x] 4) `var u_phpbb_ads_click = '{{ UA_PHPBB_ADS_CLICK }}';`
Variables with the A prefix are ment for addslashed variables, but this variable is not addslashed. It is also not escaped properly for it to be used in javascript.

- [x] 5) I would like to see the CSRF formkey check in the admin input be moved to the admin controller, to have it clear they are both done and handled correctly.

- [x] 6) It might be a good idea to look into why phpBB decided to serve all uploaded images, like avatars and attachments by downloads/file.php instead of direct linking them. I know it was for security reasons, however I can't exactly remember why.

### Comments
**5)** I have rather moved `add_form_key` to input than breaking nicely coupled validation into two classes.

**6)** It's not applicable to us. The main problem is that it is [bound to attachments](https://github.com/phpbb/phpbb/blob/master/phpBB/download/file.php#L167), so in case admin disallows attachments, but wants to display ads, this method would effectively disallow any banner to display.

### DIFF
--- a/adm/style/event/acp_overall_footer_after.html
+++ b/adm/style/event/acp_overall_footer_after.html
@@ -36,7 +36,7 @@
 			minDate: 0,
 			timepicker: false
 		});
-		$.datetimepicker.setLocale('{{ S_USER_LANG|e('js') }}');
+		$.datetimepicker.setLocale('{{ S_USER_LANG }}');
 
 		$('#upload_banner').click(function(e) {
 			e.preventDefault();
--- a/adm/style/event/acp_overall_footer_after.html
+++ b/adm/style/event/acp_overall_footer_after.html
@@ -36,7 +36,7 @@
 			minDate: 0,
 			timepicker: false
 		});
-		$.datetimepicker.setLocale('{{ S_USER_LANG }}');
+		$.datetimepicker.setLocale('{{ S_USER_LANG|e('js') }}');
 
 		$('#upload_banner').click(function(e) {
 			e.preventDefault();
@@ -70,7 +70,7 @@
 				});
 			}
 			else {
-				phpbb.alert('{{ lang('INFORMATION') }}', '{{ lang('NO_FILE_SELECTED') }}');
+				phpbb.alert('{{ lang('INFORMATION')|e('js') }}', '{{ lang('NO_FILE_SELECTED')|e('js') }}');
 			}
 		});
 	});
--- a/controller/admin_controller.php
+++ b/controller/admin_controller.php
@@ -186,7 +186,7 @@ class admin_controller
 		$upload_banner = $this->request->is_set_post('upload_banner');
 		$analyse_ad_code = $this->request->is_set_post('analyse_ad_code');
 
-		add_form_key('phpbb/ads/add');
+		$this->input->add_form_key('phpbb/ads/add');
 		if ($preview || $submit || $upload_banner || $analyse_ad_code)
 		{
 			$data = $this->input->get_form_data('phpbb/ads/add');
@@ -245,7 +245,7 @@ class admin_controller
 		$upload_banner = $this->request->is_set_post('upload_banner');
 		$analyse_ad_code = $this->request->is_set_post('analyse_ad_code');
 
-		add_form_key('phpbb/ads/edit/' . $ad_id);
+		$this->input->add_form_key('phpbb/ads/edit/' . $ad_id);
 		if ($preview || $submit || $upload_banner || $analyse_ad_code)
 		{
 			$data = $this->input->get_form_data('phpbb/ads/edit/' . $ad_id);
--- a/controller/admin_controller.php
+++ b/controller/admin_controller.php
@@ -117,11 +117,10 @@ class admin_controller
 	{
 		$this->setup();
 
-		add_form_key('phpbb/ads/settings');
 		if ($this->request->is_set_post('submit'))
 		{
 			// Validate form key
-			if (check_form_key('phpbb/ads/settings'))
+			if (check_form_key('phpbb_ads'))
 			{
 				$this->config->set('phpbb_ads_adblocker_message', $this->request->variable('adblocker_message', 0));
 				$this->config->set('phpbb_ads_enable_views', $this->request->variable('enable_views', 0));
@@ -186,10 +185,9 @@ class admin_controller
 		$upload_banner = $this->request->is_set_post('upload_banner');
 		$analyse_ad_code = $this->request->is_set_post('analyse_ad_code');
 
-		$this->input->add_form_key('phpbb/ads/add');
 		if ($preview || $submit || $upload_banner || $analyse_ad_code)
 		{
-			$data = $this->input->get_form_data('phpbb/ads/add');
+			$data = $this->input->get_form_data();
 
 			if ($preview)
 			{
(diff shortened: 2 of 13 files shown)


adm/style/event/acp_overall_footer_after.html
```diff
@@ -36,7 +36,7 @@
 			minDate: 0,
 			timepicker: false
 		});
-		$.datetimepicker.setLocale('{{ S_USER_LANG }}');
+		$.datetimepicker.setLocale('{{ S_USER_LANG|e('js') }}');
```
VSEphpbb: This should not be escaped. This is an argument for the function.

event/main_listener.php
```diff
@@ -122,7 +122,7 @@ public function setup_ads()
 					'S_INCREMENT_VIEWS'		=> true,
 					// Obfuscate URL to prevent crawlers increasing view counters.
 					// Uses http://www.jsfuck.com/ to make 'a' really complicated, yet executable.
-					'UA_PHPBB_ADS_VIEWS'	=> str_replace('adsview', "' + (![]+[])[+!+[]] + 'dsview", $this->controller_helper->route('phpbb_ads_view', array('data' => implode('-', $ad_ids)))),
+					'U_PHPBB_ADS_VIEWS'	=> str_replace('adsview', "' + (![]+[])[+!+[]] + 'dsview", $this->controller_helper->route('phpbb_ads_view', array('data' => implode('-', $ad_ids)))),
```
VSEphpbb: Because this is now being JS escaped you probably do not need the obfuscation trick here any more. The escaped output in HTML looks like:
```
$.get('\x2FphpBB\x2Fapp.php\x2F\x27\x20\x2B\x20\x28\x21\x5B\x5D\x2B\x5B\x5D\x29\x5B\x2B\x21\x2B\x5B\x5D\x5D\x20\x2B\x20\x27dsview\x2F1');
```
Anyways, the escaping is breaking the obfuscation.

The unobfuscated version will look like this:
```
$.get('\x2FphpBB\x2Fapp.php\x2Fadsview\x2F1');
```
So we can't have both obfuscation and escaping here.

styles/all/template/includes/phpbb_ads_pop_up.html
```diff
@@ -3,8 +3,8 @@
 		'use strict';
 
 		$(window).on('load', function() {
-			document.cookie = '{{ POP_UP_COOKIE_NAME }}=true; expires={{ POP_UP_COOKIE_EXPIRES }}; path={{ POP_UP_COOKIE_PATH }}';
-			phpbb.alert('{{ lang('ADVERTISEMENT')|upper }}', `{% include '@phpbb_ads/phpbb_ads_default.html' %}`);
+			document.cookie = '{{ POP_UP_COOKIE_NAME|e('js') }}=true; expires={{ POP_UP_COOKIE_EXPIRES|e('js') }}; path={{ POP_UP_COOKIE_PATH|e('js') }}';
```
VSEphpbb: These cookie settings should not be escaped. They are values so escaping them breaks them.

Example: https://github.com/phpbb/phpbb/blob/master/phpBB/styles/prosilver/template/ucp_register.html#L10

controller/admin_controller.php
```diff
@@ -186,7 +186,7 @@ public function action_add()
 		$upload_banner = $this->request->is_set_post('upload_banner');
 		$analyse_ad_code = $this->request->is_set_post('analyse_ad_code');
 
-		add_form_key('phpbb/ads/add');
+		$this->input->add_form_key('phpbb/ads/add');
```
VSEphpbb: This is not what @paul999  was asking for, and this doesn't really do any good (it's zombie code). He was asking for `check_form_key()` to be done in the admin controller. However, that could actually bust up our design since we've moved all our form data handling to the input class. To move that check here, would make it things alit more confused as we'd have to get an error message sent back to the error property of the input class. So I think that's a validation request we can ignore...especially considering we still plan to improve our form input/request handling soon.

VSEphpbb:
@senky I think for item 6, @paul999 meant maybe we should consider using the download/file.php to serve the images that we allow to be uploaded. I do not think this is necessary, because part of the reason for that is how phpBB tracks download counts of attachments, for example. We're just serving images and we already strictly limit what can be uploaded and served to only image types, which the upload class checks for us.

controller/admin_controller.php
senky: @paul999 wanted form key assignment and check in the same class so that it's easily comprehended. He suggested to move `check_form_key()` to admin controller as a solution. I did the opposite - moved `add_form_key()` to the input class which does all the validation stuff. So I think this solution resolves Paul's request and yet maintains our classes in good condition regarding number of concerns.

senky:
@VSEphpbb I got what Paul meant. I just did a little research and realised it checks for some config values that are not applicable to us and may disable displaying of banners. So you are right - it's not necessary and it's not even applicable due to config stuff.

controller/admin_controller.php
VSEphpbb: I don't think so. Because add key is still being done in the admin controller. All that's in the input class is a call to the function. So visually in the controller you see the add for key called, but never checked.

I still think this can be ignored for now, as we should improve the input class in general.

paul999: What I ment is that the add and the check should be as close as possible together, so you directly see that the checks are done. It is not something deny worthy, just a suggestion to make validation easier, and the chance of errors smaller.

paul999:
No, that download/file.php was not added for counters and stuff, but was added for security reasons, but I can't remember exactly why. I am sure it is somewhere documented from < 2007, I guess in CVS commit logs. I also don't think you should specific use download/file.php, a controller that serves it is also fine by me :P. (Also, I can see why users actually would love to see stats about how often a ad is shown)

controller/admin_controller.php
VSEphpbb: @senky I would maybe move add_form_key() into the input setup method (it does not need to be unique for add vs edit IMO, and can just use a generic name for our ads page). At least that way, both add_form_key and check_form_key will be in the same class.

VSEphpbb: Oh I thought the setup function was in input. never mind that then.

senky: So what now? Revert all changes about this, or keep it as it is or add `add_form_key()` to input's constructor?

VSEphpbb: Yeah revert these changes regarding add_form_key for now.

VSEphpbb: Well, one thing you can do is call `add_form_key('phpbb_ads')` in the inout class constructor.

Then be sure to remove all 3 add_form_key calls in the admin_controller and change the argument in check_form_key to `check_form_key('phpbb_ads')`

PR #89, opened 2017-08-31
Use the user loader

### LINKED ISSUE #85: SQL Error with Emoji characters
Trying to submit an ad with emoji resulted in this fatal error. If we can't support them, maybe we should see if we can gracefully handle the error.
```
User Error: SQL ERROR [ mysqli ]<br /><br />Incorrect string value: '\xF0\x9F\x98\x80&l...' for column 'ad_code' at row 1 [1366]<br /><br />SQL<br /><br />UPDATE phpbb_ads
SET ad_name = 'My Ad 1', ad_note = '', ad_code = '&amp;lt;a href=&amp;quot;#&amp;quot;&amp;gt;TEST AD 😀&amp;lt;/a&amp;gt;', ad_enabled = 0, ad_end_date = 0, ad_priority = 5, ad_views_limit = 0, ad_clicks_limit = 0, ad_owner = '2'
WHERE ad_id = 1
```

### LINKED ISSUE #87: What is this code?
I don't understand what the point of this code is, in the prepare_ad_owner function:

```php

if (empty($user_name))
{
    return $user_id[0];
}

```
It's not able to be reached in our tests, and I can't figure out under what cases this would be needed nor can I tell if what it's returning is even correct?

Also wouldn't it be better to just use `user_loader->get_username()` instead of this awkward `user_get_id_name()` function throughout this extension?

VSEphpbb:
Closes #87 
Fixes #85 
@senky 

mainly because @paul999 threw up when he saw `user_get_id_name` and I don't really like that function either.

This change also led me to refactor the validators, they now return the value being validated. This is done, in general, so that the date and ad owner validators can return the properly modified new values themselves (so ad owner is converted from username to id if valid, and date is converted to timestamp if valid).

One UX change resulting from this, if an ad owner was deleted, when you edit their ad, you will now see "Anonymous" listed under ad owner, which makes some sense since the ad is still assigned to a user that is gone. But you can't save it with Anonymous, attempting to will notify you to fix the ad owner.

Updating the validators also led me to add a validator for the Ad Code, to test for emoji characters which cause a fatal SQL error.

Also this PR includes some fixes to tests that were discovered.

### DIFF
--- a/controller/admin_input.php
+++ b/controller/admin_input.php
@@ -20,6 +20,9 @@ class admin_input
 	/** @var \phpbb\user */
 	protected $user;
 
+	/** @var \phpbb\user_loader */
+	protected $user_loader;
+
 	/** @var \phpbb\language\language */
 	protected $language;
 
@@ -35,14 +38,16 @@ class admin_input
 	/**
 	 * Constructor
 	 *
-	 * @param \phpbb\user								$user			User object
-	 * @param \phpbb\language\language                  $language       Language object
-	 * @param \phpbb\request\request					$request		Request object
-	 * @param \phpbb\ads\banner\banner					$banner			Banner upload object
+	 * @param \phpbb\user              $user        User object
+	 * @param \phpbb\user_loader       $user_loader User loader object
+	 * @param \phpbb\language\language $language    Language object
+	 * @param \phpbb\request\request   $request     Request object
+	 * @param \phpbb\ads\banner\banner $banner      Banner upload object
 	 */
-	public function __construct(\phpbb\user $user, \phpbb\language\language $language, \phpbb\request\request $request, \phpbb\ads\banner\banner $banner)
+	public function __construct(\phpbb\user $user, \phpbb\user_loader $user_loader, \phpbb\language\language $language, \phpbb\request\request $request, \phpbb\ads\banner\banner $banner)
 	{
 		$this->user = $user;
+		$this->user_loader = $user_loader;
 		$this->language = $language;
 		$this->request = $request;
 		$this->banner = $banner;
@@ -97,7 +102,7 @@ class admin_input
 		}
 
 		// Validate each property. Every method adds errors directly to $this->errors.
-		foreach ($data as $prop_name => $prop_val)
+		foreach ($data as $prop_name => &$prop_val)
 		{
 			if (method_exists($this, 'validate_' . $prop_name))
 			{
@@ -105,12 +110,11 @@ class admin_input
 			}
 		}
 
+		unset($prop_val);
+
 		// Replace end date and owner with IDs that will be stored in the DB
 		$data['ad_end_date'] = $this->end_date_to_timestamp($data['ad_end_date']);
-		if (!in_array('AD_OWNER_INVALID', $this->errors))
-		{
-			$data['ad_owner'] = $this->owner_to_id($data['ad_owner']);
-		}
+		$data['ad_owner'] = $data['ad_owner'] === ANONYMOUS ? 0 : (int) $data['ad_owner'];
 
 		return $data;
 	}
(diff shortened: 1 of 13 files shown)


controller/admin_input.php
```diff
@@ -93,24 +98,20 @@ public function get_form_data()
...
 		if (!check_form_key('phpbb_ads'))
 		{
-			$this->errors[] = $this->language->lang('FORM_INVALID');
+			$this->errors[] = 'FORM_INVALID';
 		}
 
-		// Validate each property. Every method adds errors directly to $this->errors.
-		foreach ($data as $prop_name => $prop_val)
+		// Validate each property. Some validators update the property value. Errors are added to $this->errors.
+		foreach ($data as $prop_name => &$prop_val)
```
senky: To be honest, I really really don't like this referencing. It's difficult to read and comprehend. I would like to see validation and data transformation separated, because they do different tasks. I know that current architecture makes this method look more like a procedure with different concerns, but at least it's readable. Proper solution would be creating a class for every form field with `validate` and `prepare_to_storage` methods.

controller/admin_input.php
```diff
@@ -156,31 +157,54 @@ public function banner_upload($ad_code)
...
+
+	/**
+	 * Validate advertisement code
+	 *
+	 * @param string $ad_code Advertisement code
+	 * @return string Advertisement code
+	 */
+	protected function validate_ad_code($ad_code)
+	{
+		if (preg_match_all('/[\x{10000}-\x{10FFFF}]/u', $ad_code, $matches))
```
senky: Can't we somehow change DB column type to support emoji? I can imagine many people attempting to use them in their ad codes...

controller/admin_input.php
```diff
@@ -189,87 +213,73 @@ protected function validate_ad_end_date($end_date)
...
-	 * @param string $ad_owner Advertisement owner
+	 * @param string $ad_owner User name
+	 * @return int User id if user exists, otherwise 0.
 	 */
 	protected function validate_ad_owner($ad_owner)
 	{
-		// user_get_id_name function returns false if everything is OK.
-		if (!empty($ad_owner) && user_get_id_name($ad_owner_id, $ad_owner))
+		$user_id = 0;
+		if (!empty($ad_owner) && ANONYMOUS === ($user_id = (int) $this->user_loader->load_user_by_username($ad_owner)))
```
senky: Please, put `$user_id = (int) $this->user_loader->load_user_by_username($ad_owner)` on separate line and remove `$user_id = 0;` (it will never be used, because it's always reassigned with `user_loader:: load_user_by_username` call.

VSEphpbb: I did it this way in order to avoid unnecessary calls to `load_user_by_username`. I don't want it being called if the Ad owner field is blank. So that's why user_id is initialized to 0, so that if the field is empty we return 0. Also if it is ANONYMOUS (meaning user didnt exist) we return 0 too.

controller/admin_input.php
VSEphpbb: Changing the architecture like you mention is good idea but a bigger job for another issue/PR.

controller/admin_input.php
VSEphpbb: Not that I'm aware of. The way phpBB handles them is the BBCode engine converts them toa  built in library of emoji using it's own thing like html entities, not the regular 4-byte. To change the DB itself requires changing the tables character encoding and the whole database collation mode, which are not things we should be changing.

controller/admin_input.php
senky: Makes sense. It's just too difficult to read. What about this?
```
$user_id = 0;
if (!empty($ad_owner))
{
	$user_id = (int) $this->user_loader->load_user_by_username($ad_owner);
	if ($user_id === ANONYMOUS)
	{
		$this->errors[] = 'AD_OWNER_INVALID';
	}
}
```

VSEphpbb: Too many nested ifs. That's something I don't like. I honestly don't find it hard to read 😄 I added more explaning to the comments.

Also this change is doing some other fixes, like changing $ad_owner. It was taking in a variable as one type (string) and then change it to some new type (int) on the way out. It's better to not change the original input variable here, and instead generate a new variable holding the new result and type for the return.

PR #93, opened 2017-09-09
Fix migrations again for delete data

VSEphpbb:
@senky

### DIFF
--- a/tests/functional/zbase_test.php
+++ b/tests/functional/zbase_test.php
@@ -1,0 +1,55 @@
+<?php
+/**
+ *
+ * Advertisement management. An extension for the phpBB Forum Software package.
+ *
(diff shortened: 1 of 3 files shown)


tests/functional/zbase_test.php
```diff
@@ -0,0 +1,55 @@
...
+ * @license GNU General Public License, version 2 (GPL-2.0)
+ *
+ */
+
+namespace phpbb\ads\tests\functional;
+
+/**
+ * @group functional
+ */
+class zbase_test extends \phpbb_functional_test_case
```
senky: Why `zbase_test`?

VSEphpbb: to ensure it is last because it deletes the extension

senky: Right! Can you, please, add this explanation to some comment in the file?

PR #100, opened 2017-10-03
UCP Ad Owner Improvements

### LINKED ISSUE #83: UCP Ad Owner Improvements
There are some extra improvements we should make to the UCP Ads page.

1. Display the End date in a column (if there is one).
2. Display click and view **limits** like we do in the ACP.
3. Enabled/Disabled/Expired status should be indicated in some way (for example like we do in the ACP with active and expired ads separation).

This way the ad owner better knows when their ad is scheduled to expire, or if it is already expired due to the date, clicks or views, and they should know if it is enabled or not.

senky:
Resolves #83


controller/helper.php
```diff
@@ -13,7 +13,7 @@
 /**
  * Admin helper
```
VSEphpbb: Just name it Helper here too

language/en/ucp.php
```diff
@@ -19,8 +19,13 @@
...
-	'NO_ADS'	=> '<strong>No advertisements to display. Users who own advertisements displayed on this board can view their statistics here.</strong>',
+	'AD_NAME'		=> 'Name',
+	'AD_END_DATE'	=> 'End Date',
+	'AD_VIEWS'		=> 'Views',
+	'AD_CLICKS'		=> 'Clicks',
+	'AD_STATUS'		=> 'Status',
+	'EXPIRED'		=> 'Expired',
+	'ACTIVE_ADS'	=> 'Active ads',
+	'EXPIRED_ADS'	=> 'Expired ads',
+	'NO_ADS'		=> '<strong>No advertisements to display. Users who own advertisements displayed on this board can view their statistics here.</strong>',
```
VSEphpbb: I still don't like this line. I suggest changing the tab from "Advertisements" to "My advertisements" and this text to: "You do not have any advertisements displayed on this board."

styles/prosilver/template/ucp_ads_stats.html
```diff
@@ -7,19 +7,62 @@
...
+							{% endif %}
+							{% if S_CLICKS_ENABLED %}
+								<td class="center">
+									{% if ad.CLICKS_LIMIT and ad.CLICKS >= ad.CLICKS_LIMIT %}<strong class="error">{% endif %}
+										{{ ad.CLICKS }}{% if ad.CLICKS_LIMIT %} / {{ ad.CLICKS_LIMIT }}{% endif %}
+										{% if ad.CLICKS_LIMIT and ad.CLICKS >= ad.CLICKS_LIMIT %}</strong>{% endif %}
+								</td>
+							{% endif %}
+							<td class="center">
+								{% if list.loop == loops.expired %}
```
VSEphpbb: Should be `if ad == loops.expired` I think

senky: Nope, it is true when current loop is "expired", not when current ad belongs to expired loop. Current solution works (tested manually).

PR #101, opened 2017-10-03
Make ad container responsive

### LINKED ISSUE #97: Unworkable ad code
There's a pair of examples where the code given by Google and Yandex won't be shown. The extension code analyzer reports the code is ok.
```
<script async src="//pagead2.googlesyndication.com/pagead/js/adsbygoogle.js"></script>
<!-- 2017.08.01 -->
<ins class="adsbygoogle"
     style="display:block"
     data-ad-client="ca-pub-6052584654693494"
     data-ad-slot="5495024367"
     data-ad-format="auto"></ins>
<script>
(adsbygoogle = window.adsbygoogle || []).push({});
</script>
```

```
<!-- Yandex.RTB R-A-161344-1 -->
<div id="yandex_rtb_R-A-161344-1"></div>
<script type="text/javascript">
    (function(w, d, n, s, t) {
        w[n] = w[n] || [];
        w[n].push(function() {
            Ya.Context.AdvManager.render({
                blockId: "R-A-161344-1",
                renderTo: "yandex_rtb_R-A-161344-1",
                async: true
            });
        });
        t = d.getElementsByTagName("script")[0];
        s = d.createElement("script");
        s.type = "text/javascript";
        s.src = "//an.yandex.ru/system/context.js";
        s.async = true;
        t.parentNode.insertBefore(s, t);
    })(this, this.document, "yandexContextAsyncCallbacks");
</script>
```

### LINKED ISSUE #98: Responsive ad container
We should probably make our ad container responsive.

https://www.phpbb.com/customise/db/extension/ads/support/topic/183086?p=589971#p589971

senky:
... and ready for Google Ads

Resolves #97 
Resolves #98

### DIFF
--- a/controller/admin_input.php
+++ b/controller/admin_input.php
@@ -127,7 +127,7 @@ class admin_input
 			$this->banner->create_storage_dir();
 			$realname = $this->banner->upload();
 
-			$banner_html = '<img class="phpbb-ads-banner" src="' . generate_board_url() . '/images/phpbb_ads/' . $realname . '" />';
+			$banner_html = '<img src="' . generate_board_url() . '/images/phpbb_ads/' . $realname . '" />';
 
 			if ($this->request->is_ajax())
 			{
--- a/controller/admin_input.php
+++ b/controller/admin_input.php
@@ -127,7 +127,7 @@ class admin_input
 			$this->banner->create_storage_dir();
 			$realname = $this->banner->upload();
 
-			$banner_html = '<div class="phpbb-ads-banner"><img src="' . generate_board_url() . '/images/phpbb_ads/' . $realname . '" /></div>';
+			$banner_html = '<img class="phpbb-ads-banner" src="' . generate_board_url() . '/images/phpbb_ads/' . $realname . '" />';
 
 			if ($this->request->is_ajax())
 			{
--- a/controller/admin_input.php
+++ b/controller/admin_input.php
@@ -127,7 +127,7 @@ class admin_input
 			$this->banner->create_storage_dir();
 			$realname = $this->banner->upload();
 
-			$banner_html = '<img src="' . generate_board_url() . '/images/phpbb_ads/' . $realname . '" />';
+			$banner_html = '<div class="phpbb-ads-banner"><img src="' . generate_board_url() . '/images/phpbb_ads/' . $realname . '" /></div>';
 
 			if ($this->request->is_ajax())
 			{
--- a/styles/all/theme/phpbbads.css
+++ b/styles/all/theme/phpbbads.css
@@ -1,6 +1,5 @@
 .phpbb-ads-center {
 	text-align: center;
-
 }
 
 .phpbb-ads-center > :first-child {
--- a/styles/all/theme/phpbbads.css
+++ b/styles/all/theme/phpbbads.css
@@ -1,6 +1,8 @@
-.phpbb-ads-center {
-	display: flex;
-	justify-content: center;
+.phpbb-ads-banner {
+	text-align: center;
+}
+.phpbb-ads-banner img {
+	max-width: 100%;
 }
 
 .phpbbad-slide-up {
--- a/styles/all/theme/phpbbads.css
+++ b/styles/all/theme/phpbbads.css
@@ -1,7 +1,10 @@
 .phpbb-ads-center {
 	text-align: center;
+
 }
-.phpbb-ads-center .phpbb-ads-banner {
+
+.phpbb-ads-center > :first-child {
+	display: inline-block;
 	max-width: 100%;
 }
 
--- a/styles/all/theme/phpbbads.css
+++ b/styles/all/theme/phpbbads.css
@@ -1,7 +1,7 @@
-.phpbb-ads-banner {
+.phpbb-ads-center {
 	text-align: center;
 }
-.phpbb-ads-banner img {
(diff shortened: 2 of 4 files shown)


styles/all/theme/phpbbads.css
```diff
@@ -1,6 +1,8 @@
 .phpbb-ads-center {
-	display: flex;
-	justify-content: center;
+	text-align: center;
```
rxu: What if we use
```css
.phpbb-ads-center > * {
	margin: 0 auto;
}
```

VSEphpbb:
What about responsive, as in ads that extend beyond the parent div? Maybe we need media queries in CSS for that?

styles/all/theme/phpbbads.css
```diff
@@ -1,6 +1,8 @@
...
+}
+.phpbb-ads-center .phpbb-ads-banner {
+	max-width: 100%;
```
VSEphpbb: wouldn't this be the same as just adding max-width 100% to the phpbb-ads-center rule?

senky:
> What about responsive, as in ads that extend beyond the parent containers? Maybe we need media queries in CSS for that?

What do you mean by that? AdSense responsive ads will deal with responsiveness itself. Do you have another use case in mind that is not covered by current solution?

styles/all/theme/phpbbads.css
senky: Nope. `max-width: 100%` must be defined on `img` tag to nicely shrink on smaller screens. And I wanted to narrow it down to our banners only, so that other ad providers can define the behaviour themselves (like with `srcset` or so).

VSEphpbb:
Ads that are not Google Responsive are not centered anymore (although the text inside them is)
screen shot 2017-10-04 at 9 19 59 am

And here is an example where a normal ad is breaking through the layout and not being handled responsively, if we even can:
screen shot 2017-10-04 at 9 21 42 am

VSEphpbb:
Maybe adding this helps (seems to center all my tests)
```css
.phpbb-ads-center * {
	display: inline-block;
}
```

styles/all/theme/phpbbads.css
VSEphpbb: OK based on my testing the special banner class is not needed. I seem to have solved all the issues with just adding this css, which applies to all first child elements of our ad container.:

```css
.phpbb-ads-center > :first-child {
	display: inline-block;
	max-width: 100%;
}
```
This will handle making banner images uploaded with the 100% max width. The inline block helps with centering divs based content. And the 100% max width also keeps all ads fitting within the page in the various responsive modes.

Some screen shots:
screen shot 2017-10-07 at 9 28 08 am

screen shot 2017-10-07 at 9 28 41 am

Full CSS change:
```css
.phpbb-ads-center {
	text-align: center;

}

.phpbb-ads-center > :first-child {
	display: inline-block;
	max-width: 100%;
}
```

rxu:
> :first-child

What if some ad have some wrapper-element aroung?

VSEphpbb:
Shouldn't matter. My test ad did. Ultimately though no matter what, some users will have to stylize their ad code if it doesn't work immediately as intended out of the box. And there's nothing that can be done about that. But we can minimize that by automatically making the centered and responsive compatible.

rxu:
I'm not sure anyway. What if you add several ads into the same location? Then you got 1st one centered but others aren't, don't you? 😃 That makes very low sense for me. Couldn't we do something like
```css
.phpbb-ads-center > * {
	display: inline-block;
	max-width: 100%;
}
```
or will it brake things?

VSEphpbb:
That would be an invalid use. Each ad should be it's own ad. Putting several into the same location would be fine because each is contained within its own phpbb-ads-center class. So they're all treated the same.

VSEphpbb:
An d no, `> * ` will break things because it applies to everything

rxu:
Ah, ok then. Thanks.

senky:
@VSEphpbb thanks for your analysis! Let's go with it and see if there are any more user complains.

controller/admin_input.php
```diff
@@ -127,7 +127,7 @@ public function banner_upload($ad_code)
 			$this->banner->create_storage_dir();
 			$realname = $this->banner->upload();
 
-			$banner_html = '<img src="' . generate_board_url() . '/images/phpbb_ads/' . $realname . '" />';
+			$banner_html = '<img class="phpbb-ads-banner" src="' . generate_board_url() . '/images/phpbb_ads/' . $realname . '" />';
```
VSEphpbb: This is no longer needed

VSEphpbb: But why would just the banner upload have this, if useful? Considering that users can paste anything they want into their ad, why is just the upload gonna get a special class? Also, this isn't anything hard-coded, it's just pasted into the code space where they can and should, manipulate it it to suit their needs, so they can add anything they want to it themselves.

VSEphpbb: I think it might actually confuse people, who have no idea if it is required or not to be in the banner code 😉

styles/all/theme/phpbbads.css
```diff
@@ -1,6 +1,11 @@
 .phpbb-ads-center {
-	display: flex;
-	justify-content: center;
+	text-align: center;
+
```
VSEphpbb: Please remove this extra line

PR #108, opened 2018-01-03
#103: Hide UCP module for members without ads

### LINKED ISSUE #103: Hide UCP module for members without ads
Just thinking out loud... Maybe there is a way to dynamically set `u_` permission? That way we could set module auth to `ext_phpbb/ads && u_ads`. And set `u_ads` to true when user has at least one ad assigned.

senky:
Resolves #103 

TODO:
- [x] Feature
- [x] Fix failing tests
- [x] Write new tests

### DIFF
--- a/controller/admin_controller.php
+++ b/controller/admin_controller.php
@@ -90,7 +90,7 @@ class admin_controller
 
 		$this->template->assign_var('S_PHPBB_ADS', true);
 
-		if (!class_exists('\auth_admin'))
+		if (!class_exists('auth_admin'))
 		{
 			include($root_path . 'includes/acp/auth.' . $php_ext);
 		}
@@ -285,7 +285,7 @@ class admin_controller
 				$this->manager->delete_ad_locations($ad_id);
 				$success = $this->manager->delete_ad($ad_id);
 
-				$this->try_remove_permission($ad_data['ad_owner']);
+				$this->toggle_permission($ad_data['ad_owner']);
 
 				// Only notify user on error or if not ajax
 				if (!$success)
@@ -459,7 +459,7 @@ class admin_controller
 		if (!$this->input->has_errors())
 		{
 			$ad_id = $this->manager->insert_ad($this->data);
-			$this->auth_admin->acl_set('user', 0, $this->data['ad_owner'], array('u_phpbb_ads' => 1));
+			$this->toggle_permission($this->data['ad_owner']);
 			$this->manager->insert_ad_locations($ad_id, $this->data['ad_locations']);
 
 			$this->helper->log('ADD', $this->data['ad_name']);
@@ -484,8 +484,8 @@ class admin_controller
 			if ($success)
 			{
 				// Only update permissions when update was successful
-				$this->try_remove_permission($old_data['ad_owner']);
-				$this->auth_admin->acl_set('user', 0, $this->data['ad_owner'], array('u_phpbb_ads' => 1));
+				$this->toggle_permission($old_data['ad_owner']);
+				$this->toggle_permission($this->data['ad_owner']);
 
 				// Only insert new ad locations to DB when ad exists
 				$this->manager->delete_ad_locations($ad_id);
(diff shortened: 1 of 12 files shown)


controller/admin_controller.php
```diff
@@ -84,6 +89,12 @@ public function __construct(\phpbb\template\template $template, \phpbb\language\
 		$this->language->add_lang('acp', 'phpbb/ads');
 
 		$this->template->assign_var('S_PHPBB_ADS', true);
+
+		if (!class_exists('\auth_admin'))
```
VSEphpbb: should not need the leading slash

controller/admin_controller.php
```diff
@@ -465,9 +479,14 @@ protected function submit_edit()
 		$ad_id = $this->request->variable('id', 0);
 		if ($ad_id && !$this->input->has_errors())
 		{
+			$old_data = $this->manager->get_ad($ad_id);
 			$success = $this->manager->update_ad($ad_id, $this->data);
 			if ($success)
 			{
+				// Only update permissions when update was successful
+				$this->try_remove_permission($old_data['ad_owner']);
+				$this->auth_admin->acl_set('user', 0, $this->data['ad_owner'], array('u_phpbb_ads' => 1));
```
VSEphpbb: I think we can just have 1 function for toggling permissions, called after delete ad, after insert ad, and here after update ad.

Some logic like:
if this user has ads and no permision, then set their permission. Else if user has no ads but has permission, then remove their permissions.

event/main_listener.php
```diff
@@ -79,6 +80,19 @@ public function __construct(\phpbb\template\template $template, \phpbb\user $use
...
+	 *
+	 * @param	\phpbb\event\data	$event	The event object
+	 * @return	void
+	 */
+	public function set_permissions($event)
+	{
+		$permissions = $event['permissions'];
+		$permissions['u_phpbb_ads'] = array('lang' => 'ACL_U_PHPBB_ADS', 'cat' => 'misc');
+		$event['permissions'] = $permissions;
+	}
```
VSEphpbb: we could try the new 1-line code here instead?

```php
$event->update_subarray('permissions', 'u_phpbb_ads', ['lang' => 'ACL_U_PHPBB_ADS', 'cat' => 'misc']);
```

language/en/permissions_ads.php
```diff
@@ -0,0 +1,23 @@
...
+	exit;
+}
+
+if (empty($lang) || !is_array($lang))
+{
+	$lang = array();
+}
+
+$lang = array_merge($lang, array(
+	'ACL_U_PHPBB_ADS' => 'Can see My advertisements UCP module',
```
VSEphpbb: Just a little English tweaking for consistency with other permissions:
"Can view own advertisement management statistics"

migrations/v10x/m12_u_phpbb_ads_permission.php
```diff
@@ -0,0 +1,84 @@
...
+		// set u_phpbb_ads to true for ad owners
+		$sql_ary = array();
+		$sql = 'SELECT ad_owner
+				FROM ' . $this->table_prefix . 'ads
+				WHERE ad_owner != 0
+				GROUP BY ad_owner';
+		$result = $this->db->sql_query($sql);
+		while ($row = $this->db->sql_fetchrow($result))
+		{
+			$sql_ary[] = array(
```
VSEphpbb: can you try using `acl_set()` here instead, to greatly simplify this whole function? The acp/auth class can be instanced here just like it was in the admin_controller.

senky:
!unset WIP

senky:
!unset needs tests

controller/admin_controller.php
```diff
@@ -500,4 +519,25 @@ protected function error($msg)
...
+	 * Permission is only removed when user has no more ads.
+	 * Permission is only added when user has at least one ad.
+	 *
+	 * @param	int	$user_id	User ID to try to remove permission
+	 *
+	 * @return	void
+	 */
+	protected function toggle_permission($user_id)
+	{
+		if (count($this->manager->get_ads_by_owner($user_id)) === 0)
```
VSEphpbb: I think this can be simplified to just:

```php
$has_ads = count($this->manager->get_ads_by_owner($user_id)) !== 0;

$this->auth_admin->acl_set('user', 0, $user_id, array('u_phpbb_ads' => (int) $has_ads));
```

migrations/v10x/m12_u_phpbb_ads_permission.php
```diff
@@ -0,0 +1,73 @@
...
+	public function set_u_phpbb_ads_permission()
+	{
+		if (!class_exists('auth_admin'))
+		{
+			include($this->phpbb_root_path . 'includes/acp/auth.' . $this->php_ext);
+		}
+		$auth_admin = new \auth_admin();
+
+		$sql = 'SELECT ad_owner
+				FROM ' . $this->table_prefix . 'ads
```
VSEphpbb: too many tabs?

PR #110, opened 2018-01-14
Simplify SQL query in get_ads

senky:
As noticed in one [support topic](https://www.phpbb.com/customise/db/extension/ads/support/topic/185951), SQL query for getting ads is overcomplicated. Simplifying also seems to fix randomness problem for [some](https://www.phpbb.com/customise/db/extension/ads/support/topic/185951?p=606311#p606311) [users](https://www.phpbb.com/customise/db/extension/ads/support/topic/185951?p=611816#p611816).

### DIFF
--- a/ad/manager.php
+++ b/ad/manager.php
@@ -69,9 +69,7 @@ class manager
 		$sql_where_views = $this->config['phpbb_ads_enable_views'] ? 'AND (a.ad_views_limit = 0 OR a.ad_views_limit > a.ad_views)' : '';
 		$sql_where_clicks = $this->config['phpbb_ads_enable_clicks'] ? 'AND (a.ad_clicks_limit = 0 OR a.ad_clicks_limit > a.ad_clicks)' : '';
 
-		$sql = 'SELECT location_id, ad_id, ad_code
-			FROM (
-				SELECT al.location_id, a.ad_id, a.ad_code
+		$sql = 'SELECT al.location_id, a.ad_id, a.ad_code
 				FROM ' . $this->ad_locations_table . ' al
 				LEFT JOIN ' . $this->ads_table . ' a
 					ON (al.ad_id = a.ad_id)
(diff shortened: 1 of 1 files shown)


senky:
@VSEphpbb I have no confidence it is better. I think it's the same. Users claim it to be better. But if you look closer, it's just so obvious result output is the same, because query is basically the same, only simpler. Old query firstly ordered all ads by random and then by location_id. Now, ads are ordered by location_id and then random. That's the only change. I just suppose it's faster because of eliminated subquery.

Leinad4Mind:
I would suggest this instead of the last order: GROUP BY al.location_id, a.ad_id, (RAND() * a.ad_priority) DESC';

Why?

I've runned the original query and it gived me this:
https://i.imgur.com/EZB4qb8.png

If I run this new query:
https://i.imgur.com/eDnW3QO.png

With my sugestion:
https://i.imgur.com/SawbHvw.png

The exact same result.
BUT
With my suggestion that gives the same result as the original code, the bug is still present! Very strange indeed!

So, yeah, no need of confidence, just facts, since it mysteriously fix the bug.

Maybe this information can help you guys out to understand why the old query was failing.

VSEphpbb:
GROUP BY was not compatible with other DBMs

Leinad4Mind:
I see.
Btw I've saw that the Senky Query everytime it runs, the output is different and thats why it works. 
The Random works. 
On the original code since the order is after the random, it will output always the same results, and therefore the random doesnt exist. Thats why people are getting always the same ads.

VSEphpbb:
Also the old query never failed for us, any of our testers and not in all our unit tests, so we actually attribute this more to some strange issues with the web servers of the 3 or 4 people who posted about it, but not with total certainty of course.

Leinad4Mind:
The only thing that makes the old query to fail is "ORDER BY z.location_id"

If we remove that Order By, it will random again.

So that order by location_id, in reality is ordering too by ad_id (it shouldn't...)

Leinad4Mind:
The reason is: The outer order by overwrites the inner order by. You cannot concatenate an order by with a previous order by. Because the inner order by will be meaningless with an outer one. 👍

PR #113, opened 2018-01-15
Hide ad on Sign-in, SignUp, Profile, MCP and UCP

senky:
https://trello.com/c/5IyXEOwh/51-hide-ad-on-sign-in-signup-profile-mcp-and-ucp-option


senky:
I think it handles pretty much everything except for error (trigger) messages, which can be displayed on any page. But can we somehow catch `trigger_error()` call and hide our ads in such case?

VSEphpbb:
This should not be a general setting. It should be an ad setting. Since it's really only an issue for Google AdSense.

Haven't looked too deep into this yet, but should be much easier to handle then just by checking like if the user is in UCP/MCP. So if ad has this setting and user is in UCP or MCP, don't show.

VSEphpbb:
I thik this can be done in another much simpler way. First I don't think we need to be concerned about error pages. But, for example:
```twig
{% if PHPBB_ADS_CODE and not S_IN_MCP and not S_IN_UCP %}
```
Of course there are other options, like how we define `PHPBB_ADS_CODE`. It can be made empty if the ad has this setting enabled for this ad and the user is currently in MCP or UCP.

I think ads are fine on search and meberlist. It's only really unneccesary in the control panels which are the truly non-content pages of the forum.

senky:
@VSEphpbb I don't know. Google's Ad placement policies describe exactly:

> Publishers are not permitted to place ads on any non-content based pages like **thank you, error, log in, or exit pages**.

If we are going to do it, let's do it properly and fully. Otherwise we might find ourselves in tons of user requests about incomplete feature...

I am OK with generalizing UCP and MCP and leaving search and contact us pages untouched, but error pages must be covered. I think it might be as easy as checking for `MESSAGE_TEXT` template variable existence. Isn't it?

VSEphpbb:
Sounds good

senky:
@VSEphpbb rewritten to be ad-specific option. I don't know why, but all tests fail as soon as I added new column to fixture. Can you tell me where is the problem?

ad/manager.php
```diff
@@ -62,12 +62,14 @@ public function get_ad($ad_id)
 	 * Get one ad per every location
 	 *
 	 * @param    array $ad_locations List of ad locations to fetch ads for
+	 * @param    bool  $non_content_page Is current page without content?
 	 * @return    array    List of ad codes for each location
 	 */
-	public function get_ads($ad_locations)
+	public function get_ads($ad_locations, $non_content_page)
```
VSEphpbb: Can we not do this? I'd prefer to have the `a.ad_hide_on_noncontent` always as part of the ad's data array and check it in the event listener. like:
```php
foreach ($this->manager->get_ads($location_id) as $row)
{
    if ($non_content_page && $row['ad_hide_on_no_content'])
    {
        continue;
    }
```
Should simplify things a little and be cleaner.

I also kinda wish the variables names here can be less long, lol.

VSEphpbb: optional argument, default to false?

Leinad4Mind:
Thanks for this. Waiting for next version so I can use AdSense by Google ;)
Great Job, Guys!

ad/manager.php
```diff
@@ -62,12 +62,14 @@ public function get_ad($ad_id)
 	 * Get one ad per every location
 	 *
 	 * @param    array $ad_locations List of ad locations to fetch ads for
+	 * @param    bool  $non_content_page Is current page without content?
```
VSEphpbb: Lets just clean up this comment: `Is current page non-content oriented (e.g.: login, UCP, MCP)? Default is false.`

ad/manager.php
```diff
@@ -336,15 +339,16 @@ public function load_groups()
...
-			'ad_views_limit'	=> '',
-			'ad_clicks_limit'	=> '',
-			'ad_owner'			=> '',
+			'ad_name'				=> '',
+			'ad_note'				=> '',
+			'ad_code'				=> '',
+			'ad_enabled'			=> '',
+			'ad_end_date'			=> '',
+			'ad_priority'			=> '',
+			'ad_hide_on_noncontent'	=> '',
```
VSEphpbb: This may be a small big change (sorry in advance) but perhaps a better name would be to flip the boolean, and rename it to `ad_content_only`.

For me it's easier to understand if it's called content only instead of noncontent and especially hide_on_noncontent. I think reading through the code it's easier to understand if it were just ad_content_only.

Likewise,the language would be changed from `Hide on non-content pages` to `Display on content pages only`. For me this makes more immediate sense and may help out with all the foreign users understanding too.

language/en/acp.php
```diff
@@ -43,8 +43,8 @@
 	'AD_LOCATIONS_EXPLAIN'		=> 'Select locations where you want this advertisement displayed. Mouse over a location for a short description of it. If multiple ads use the same location, one ad will be randomly selected to display in that location each time. Use CTRL+CLICK (or CMD+CLICK on Mac) to select/deselect more than one location.',
 	'AD_PRIORITY'				=> 'Priority',
 	'AD_PRIORITY_EXPLAIN'		=> 'Set a number between 1 and 10. Advertisements with higher number will be displayed more often when there are multiple ads using the same location.',
-	'AD_HIDE_ON_NONCONTENT'		=> 'Hide on non-content pages',
-	'AD_HIDE_ON_NONCONTENT_EXPLAIN'	=> 'This will hide this ad on non-content pages such as UCP, login, registration, etc. Some advertising platforms (e.g. Google AdSense) require this.',
+	'AD_CONTENT_ONLY'			=> 'Display on content pages only',
+	'AD_CONTENT_ONLY_EXPLAIN'	=> 'This will hide this ad on non-content pages such as UCP, login, registration, etc. Some advertising platforms (e.g. Google AdSense) require this.',
```
VSEphpbb: Lets just re-wword this to:
> This ad will only display on pages that contain content. It will not be shown on pages such as the UCP, login, registration, etc. Some advertising platforms (e.g. Google AdSense) require this.

PR #114, opened 2018-01-18
Introduce Visual demo for ad locations

### LINKED ISSUE #95: Ad Locations Visual Demo
Here' an idea:

Some clickable link to open a new window that shows a "dummy" ad in every available location of the current board, so users can get a visual idea of where the ad locations are. The dummy ad would just be a banner shaped block with the name of the ad location in it.

It would look something like in this screenshot:
snimka obrazovky 2017-08-21 o 8 29 31

This is a low priority but would be nice to have at some point, I think.

senky:
Closes #95 

First attempt here.


senky:
@VSEphpbb are you OK with this implementation?

VSEphpbb:
It's great

senky:
@VSEphpbb I can't get functional `visual_demo_test` working. Whatever method I use to access `index.php?enable_visual_demo=true`, the parameter is somehow lost on the way and thus visual demo is not activated. Can you, please, take a look at it?

VSEphpbb:
@senky Alright a couple of things. The test is failing because of cookies and redirects. So first of all I do not think you should be doing a redirect immediately after setting a cookie. We really should always try to avoid redirecting. I will add some comments in line. They may fix the issue, and if not, then you may need to look at cookie stuff in functional tests in phpBB using `cookiejar`.

event/main_listener.php
```diff
@@ -223,6 +262,25 @@ public function remove_ad_owner($event)
...
+		}
+		else if ($this->request->is_set('disable_visual_demo'))
+		{
+			$this->user->set_cookie('phpbb_ads_visual_demo', '', 1);
+
+			// disable visual demo for this request
+			$this->template->assign_var('S_PHPBB_ADS_VISUAL_DEMO', false);
+			$this->in_visual_demo = false;
+
+			trigger_error('VISUAL_DEMO_DISABLED', E_USER_NOTICE);
```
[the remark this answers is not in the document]
senky: Sure it is needed. We need to inform user that visual demo has been disabled and since we opened a new window for it (user might not be aware of this), we should tell him that he can close the window now. Also since it uses `core.page_header` now (so that `$this->in_visual_demo` can be used without redirects in other methods) phpBB won't catch the error. We really need to use `trigger_error()`.

But I added small JS snippet that closes the window automatically when user has JS enabled. So this message will be be only displayed very few times.

VSEphpbb:
I don’t see it needed in my uses. It’s pretty obvious when all the demo ads disappear, after clicking to disable it. We don’t really need to confirm this for them.

VSEphpbb:
...but if you really want a confirm message we need to find a way do it without trigger error, since trigger error is kind of a bad practice we are trying to stop using.

senky:
@VSEphpbb ok I removed it (but kept JS part). Functional test still fails. I will have to take a close look at cookies...

tests/functional/visual_demo_test.php
```diff
@@ -0,0 +1,25 @@
...
+ * @group functional
+ */
+class visual_demo_test extends functional_base
+{
+	public function test_visual_demo()
+	{
+		$crawler = self::request('GET', "index.php?enable_visual_demo=true&sid={$this->sid}");
+
+		// We should be on index page now. Visual demo disable prompt should be displayed.
+		$this->assertContainsLang('DISABLE_VISUAL_DEMO', $crawler->filter('.rules')->text());
```
VSEphpbb: You need to include the common lang file in the functional base class. Also just check for lang key `VISUAL_DEMO` instead since the one you're using here will not work because it requires arguments for `sprintf`

language/en/acp.php
```diff
@@ -40,7 +40,7 @@
 	'BANNER_UPLOAD'				=> 'Upload banner',
 	'AD_PLACEMENT'				=> 'Advertisement placement',
 	'AD_LOCATIONS'				=> 'Locations',
-	'AD_LOCATIONS_EXPLAIN'		=> 'Select locations where you want this advertisement displayed. Mouse over a location for a short description of it. If multiple ads use the same location, one ad will be randomly selected to display in that location each time. Use CTRL+CLICK (or CMD+CLICK on Mac) to select/deselect more than one location.',
+	'AD_LOCATIONS_EXPLAIN'		=> 'Select locations where you want this advertisement displayed. Mouse over a location for a short description of it. If multiple ads use the same location, one ad will be randomly selected to display in that location each time. Use CTRL+CLICK (or CMD+CLICK on Mac) to select/deselect more than one location.<br><br>You can see visual demo of ad locations while browsing board: <a class="button2" href="%s" target="_blank">Start visual demo of ad locations</a>',
```
VSEphpbb: I still think the button can be improved. It's complicated because it's not really a button but an anchor link. 

But first, I don't think we need any of this extra language here. I would revert this line.

Then let's just have the button link simply say: `Ad Locations Visual Demo`

We can even add the font icon `fa-eye` so it looks nice. It should either be a plain URL link put into the HTML template, like: 
```twig
<span>{{ lang('AD_LOCATIONS_EXPLAIN') }}<br><br>
<a href="{U_ENABLE_VISUAL_DEMO}" target="_blank">
    <i class="fa fa-fw fa-eye"></i> <span>{{ lang('AD_LOCATIONS_VISUAL_DEMO') }}
</a></span>
```
Or we can add extra styling to that link so that it winds up looking exactly like our other buttons.

event/main_listener.php
```diff
@@ -125,7 +142,18 @@ public function load_language_on_setup($event)
 	 */
 	public function setup_ads()
 	{
-		if ($this->can_view_ads())
+		if ($this->in_visual_demo)
+		{
+			$all_locations = $this->location_manager->get_all_locations(false);
```
VSEphpbb: It seems like all this can actually be put inside the `visual_demo()` method and move where visual_demo is called to the same event as setup_ads `core.page_footer_after` and it all still works fine, and the code is more together

event/main_listener.php
```diff
@@ -177,6 +205,17 @@ public function clicks()
 		}
 	}
 
+	public function visual_demo()
```
VSEphpbb: Missing docblock

event/main_listener.php
```diff
@@ -125,7 +142,18 @@ public function load_language_on_setup($event)
...
 	{
-		if ($this->can_view_ads())
+		if ($this->in_visual_demo)
+		{
+			$all_locations = $this->location_manager->get_all_locations(false);
+			foreach ($this->location_manager->get_all_location_ids() as $location_id)
+			{
+				$this->template->assign_vars(array(
+					'AD_' . strtoupper($location_id) . '_ID'	=> $location_id,
+					'AD_' . strtoupper($location_id)			=> '<div style="background-color:red;width:200px;height:50px;color:white;line-height:50px;" title="' . $all_locations[$location_id]['desc'] . '">' . $all_locations[$location_id]['name'] . '</div>',
```
VSEphpbb: Just a personal taste idea, but maybe make these demo banners 100% width so the user sees how much real-estate their ads can occupy in the demo?

Also, probably put all this styling into our CSS stylesheet as class like `.phpbb-ads-visual-demo`

event/main_listener.php
```diff
@@ -223,6 +262,24 @@ public function remove_ad_owner($event)
 		$this->manager->remove_ad_owner($event['user_ids']);
 	}
 
+	public function manage_visual_demo()
```
VSEphpbb: Missing docblock

adm/style/manage_ads.html
```diff
@@ -78,7 +78,13 @@ <h3>{{ lang('WARNING') }}</h3>
 		<fieldset>
 			<legend>{{ lang('AD_PLACEMENT') }}</legend>
 			<dl>
-				<dt><label for="ad_locations">{{ lang('AD_LOCATIONS') ~ lang('COLON') }}</label><br /><span>{{ lang('AD_LOCATIONS_EXPLAIN') }}</span></dt>
+				<dt><label for="ad_locations">{{ lang('AD_LOCATIONS') ~ lang('COLON') }}</label><br />
+					<span>{{ lang('AD_LOCATIONS_EXPLAIN') }}<br><br>
+						<a class="phpbb-ads-button" href="{U_ENABLE_VISUAL_DEMO}" target="_blank">
+							<i class="fa fa-fw fa-eye"></i> {{ lang('AD_LOCATIONS_VISUAL_DEMO') }}
```
VSEphpbb: Wrap AD_LOCATIONS_VISUAL_DEMO in a `<span>` tag

controller/helper.php
```diff
@@ -139,6 +139,16 @@ public function get_find_username_link()
...
 	}
 
+	/**
+	 * Generate URL to enable visual demo of ad locations.
+	 *
+	 * @return	string	Enable visual demo URL
+	 */
+	public function get_enable_visual_demo_link()
+	{
+		return append_sid("{$this->root_path}index.{$this->php_ext}", 'enable_visual_demo=true');
```
VSEphpbb: I'm not a fan of `enable_visual_demo=true` and `disable_visual_demo=true`. We should probably use a simple controller instead. For example, the URL `./ads-visual-demo`

The controller can then handle all the cookie handling (see Board Announcements controller as an example).

Basically my idea is the controller will check if the cookie exists. If not, it sets the cookie and redirects to the index page. If the cookie does exist, it destroys the cookie and redirects the user to their last viewed page (or if AJAX request is detected, we use JS to close the window). So it's really just a simple cookie toggle and redirector.

This should eliminate the need for the manage_visual_demo method in the listener too. This also adds the benefit of eliminating the uneeded extra processing of always running manage visual demo on every page load in the listener.

event/main_listener.php
```diff
@@ -177,6 +194,31 @@ public function clicks()
...
 	}
 
+	/**
+	 * Generate visual demo templates
+	 *
+	 * @return	void
+	 */
+	public function visual_demo()
+	{
+		if ($this->in_visual_demo)
```
VSEphpbb: If we use the controller option, this private property probably won't be needed. We can just check if the cookie exists here and be good (since the manage visual demo method below would be obsolete.

controller/helper.php
VSEphpbb: I was bored and did a quick test. Something like this works, but it does need some extra bells and whistles for AJAX requests that will close the window instead of redirecting on disable.

```php
public function handle()
{
	// Protect against unauthorised access or CSRF/XSRF attacks
	if (!$this->auth->acl_get('a_') || !check_link_hash($this->request->variable('hash', ''), 'visual-demo'))
	{
		throw new \phpbb\exception\http_exception(403, 'NO_AUTH_OPERATION');
	}

	if ($this->request->is_set($this->config['cookie_name'] . '_phpbb_ads_visual_demo', \phpbb\request\request_interface::COOKIE))
	{
		// If cookie already exists, lets destroy it and redirect user to previous page viewed.
		$this->user->set_cookie('phpbb_ads_visual_demo', '', 1);
		$redirect = $this->request->variable('redirect', $this->user->data['session_page']);
	}
	else
	{
		// Since cookie does not exist yet, lets create one and send user to the index page.
		$this->user->set_cookie('phpbb_ads_visual_demo', '', 0);
		$redirect = './index.php';
	}

	// Do some is_ajax related magic here, return a json response so window will close via JS

	// Redirect user to a page
	$redirect = reapply_sid($redirect);
	redirect($redirect);

	// We shouldn't get here, but throw an http exception just in case
	throw new \phpbb\exception\http_exception(500, 'GENERAL_ERROR');
}

```

senky:
@VSEphpbb I don't think it's worth our time to investigate it that much. Let's just remove the tests. They are only functional after all...

PR #116, opened 2018-01-24
General updates

VSEphpbb:


VSEphpbb:
@senky more changes. I noticed the toggle permissions was being run a lot, when it did not ned to be. So fixed that.

Also fixed a PHP 7.2 error found in the tests.

PR #123, opened 2018-03-22
Make group hiding ad-wise

### LINKED ISSUE #119: Hide an ad for a group
Since this feature is [often requested](https://trello.com/c/ftuxi3FU/46-hide-specific-advertisements-for-specific-group), I am thinking how to implement it with adding as little load as possible. Still, the best solution seems to be an extra table storing relations of ads and groups:
```
phpbb_ad_group
    ad_id
    group_id
```
Query will be extended with a new `JOIN` and a new condition:
```sql
LEFT JOIN phpbb_ad_group ag ON (a.ad_id = ag.ad_id AND ag.group_id)
...
WHERE ag.group_id NOT IN (<list of user groups>)
```

Can you think of another, better solution?

senky:
Closes #119 

PR moves ability to hide ads for specific groups from overall settings to ad settings. Each ad can have different groups that will not see that specific ad.

For that purpose, new DB table has been introduced to store ad-group pairings. Note it's in `v20x` migrations directory because it breaks compatibility and we use [Semantic Versioning](https://semver.org/).

Main SQL query uses `EXISTS` instead of `JOIN` because we don't need to select the data from new table.

Adblocker friendly message is now displayed to everyone for obvious reasons.


ad/manager.php
```diff
@@ -190,14 +203,23 @@ public function insert_ad($data)
...
 
 		$sql = 'UPDATE ' . $this->ads_table . '
 			SET ' . $this->db->sql_build_array('UPDATE', $data) . '
 			WHERE ad_id = ' . (int) $ad_id;
 		$this->db->sql_query($sql);
+		$result = $this->db->sql_affectedrows();
 
-		return $this->db->sql_affectedrows();
+		// remove old ad_group table data
+		$sql = 'DELETE FROM ' . $this->ad_group_table . '
```
VSEphpbb: a `remove_ad_group_data` method would be nice instead here, to go along with the `insert_ad_group_data` one

senky: I didn't see it necessary because it's used only once...

VSEphpbb: For me it's more about consistency, and having an extra SQL query in here...in a function that's called update ad...but this query is about updating something else (group stuff) it's kind of out of place.

Also, a good idea I once learned, if you need to make a comment to explain what a chunk of code does, it's probably something that should become it's own method, so the call to the method name itself explains what is happening.

controller/admin_input.php
```diff
@@ -94,6 +94,7 @@ public function get_form_data()
 			'ad_views_limit'  	=> $this->request->variable('ad_views_limit', 0),
 			'ad_clicks_limit' 	=> $this->request->variable('ad_clicks_limit', 0),
 			'ad_owner'        	=> $this->request->variable('ad_owner', '', true),
+			'ad_groups'			=> $this->request->variable('hide_groups', array(0)),
```
VSEphpbb: can we change name from 'hide_groups' to 'ad_groups" so it matches, like all the rest of these parameters do ?

controller/admin_controller.php
```diff
@@ -202,6 +185,7 @@ protected function action_add()
 		else
 		{
 			$this->helper->assign_locations();
+			$this->helper->assign_groups();
```
VSEphpbb: Below, in the `action_edit()` method, don't you need to do the same thing for ad groups that you do for locations there too?
```
$this->data['ad_locations'] = $this->manager->get_ad_locations($ad_id);
```
It seems it would be possible there to load the ACP without the proper ad group data available.

controller/helper.php
```diff
@@ -117,6 +128,19 @@ public function assign_locations($ad_locations = false)
 		}
 	}
 
+	public function assign_groups($ad_id = 0)
```
VSEphpbb: Missing docblock

migrations/v20x/m1_hide_ad_for_group.php
```diff
@@ -0,0 +1,82 @@
...
+
+	/**
+	 * Remove phpbb_ads_hide_groups config.
+	 *
+	 * @return array Array of data update instructions
+	 */
+	public function update_data()
+	{
+		return array(
+			array('config_text.remove', array('phpbb_ads_hide_groups')),
```
VSEphpbb: Before we dump this config, we should probably save existing users some headache, by running a custom function that will check this value first, and set any groups that were selected to all existing ads. That way users who upgrade will still have all their ads hidden from the groups they chose, instead of having all their ads suddenly become visible to all users again.

ad/manager.php
```diff
@@ -173,12 +181,17 @@ public function increment_ad_clicks($ad_id)
 	 */
 	public function insert_ad($data)
 	{
+		$ad_groups = $data['ad_groups'];
```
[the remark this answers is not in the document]
senky: It's important because it gets filtered by `intersect_ad_data()`. Filtered data are then passed directly to `sql_build_array()` and SQL query would trigger error `column ad_groups does not exist`.

ad/manager.php
```diff
@@ -317,12 +339,17 @@ public function load_memberships($user_id)
...
-	public function load_groups()
+	public function load_groups($ad_id)
 	{
-		$sql = 'SELECT group_id, group_name, group_type
-			FROM ' . GROUPS_TABLE . "
-			WHERE group_name <> 'BOTS'
-			ORDER BY group_name ASC";
+		$sql = 'SELECT g.group_id, g.group_name, (
+				SELECT COUNT(ad_id)
+				FROM ' . $this->ad_group_table . ' ag
```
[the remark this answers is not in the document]
senky: It has intentionally one more tab because it's subquery.

controller/admin_controller.php
senky: In `action_edit()`, it's called inside `helper::assign_data()`. Since `assign_data()` is not called in this branch,`assign_groups()` has to be called manually. Also it needs to be called, otherwise there is no way to get all group names. Selectbox would be empty.

ad/manager.php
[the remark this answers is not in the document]
VSEphpbb: Please add a comment explaining that (here and in the update ad function too).

ad/manager.php
```diff
@@ -377,4 +401,35 @@ protected function sql_random()
...
+			$sql_ary[] = array(
+				'ad_id'		=> $ad_id,
+				'group_id'	=> $group,
+			);
+		}
+		$this->db->sql_multi_insert($this->ad_group_table, $sql_ary);
+	}
+
+	/**
+	 * @param int	$ad_id	Advertisement ID
```
VSEphpbb: Add a docblock description

controller/admin_controller.php
VSEphpbb: i see now

controller/helper.php
```diff
@@ -117,6 +128,23 @@ public function assign_locations($ad_locations = false)
 		}
 	}
 
+	/**
+	 * @param int $ad_id Advertisement ID
```
VSEphpbb: Add a docblock description

PR #124, opened 2018-04-05
Fix script content displayed

senky:
Fixes https://www.phpbb.com/customise/db/extension/ads/support/topic/191151

This is very quick fix that should be released ASAP. But whole ad location styling is broken. We should rethink it and strictly define what is our responsibility and what is responsibility of users. This situation is unsustainable.

### DIFF
--- a/styles/all/theme/phpbbads.css
+++ b/styles/all/theme/phpbbads.css
@@ -2,7 +2,7 @@
 	text-align: center;
 }
 
-.phpbb-ads-center > :first-child {
+.phpbb-ads-center > :first-child:not(script) {
 	display: inline-block;
 	max-width: 100%;
 }


VSEphpbb:
It won't be getting released ASAP as the current version has been in the queue for a month so far. So this will be a while. And in that case, if this fixes one thing but breaks another, then we should not do this, but instead try to fix it here the right way. I'm pretty sure the user that reported this could just wrap their code in a `<div>` tag for now, while we work on this.

senky:
This fix doesn't break anything. It merely narrows down the selector. So instead of

> select first element in `.phpbb-ads-center`

now it's

> select first element in `.phpbb-ads-center` that is not `script`

But you are right. It won't be released anytime soon. So let's rethink everything.

To be honest, I would simply remove everything and leave styling for admins. BUT they expect at least 2 things from us:

1. When they upload banner, it should be centered and responsive.
2. When they paste AdSense code, it should be centered and responsive.

We need to resolve the points with as narrow solution as possible, so that other cases aren't affected. What I see as the main problem is that centering thing. Right now we solve it with `text-align: center;` on `.phpbb-ads-center`. It's great because it centeres both banners and AdSense. But it also centeres every other ad and it propagates down the HTML, so e.g. a interactive banner that is originally centered, but it's copy is aligned left is now in the center.

And so we are in circles...

The best solution that comes into my mind is this:

1. Really don't put any style definition on `.phpbb-ads-center` or it's children.
2. When we upload banner, don't just include `` tag, but wrap it with `<div style="text-align: center;">`.
3. When we detect AdSense code (that might be tricky, it changes from time to time), wrap it with whatever it needs (I don't have AdSense, can't test).

Ideas?

VSEphpbb:
We could add an option like:
"Force this ad to be centered"

For ads with this, they will be centered the way we currently center all ads.

Ads that do not have this option will not be forced to center, and will appear how they are coded. I'm pretty sure AdSense can be self centering. But also it means admins can control the styling of the ad themselves better too, if they need to.

PR #130, opened 2018-09-04
Introduce scripts location

senky:
As a reaction to https://www.phpbb.com/customise/db/extension/ads/support/topic/192696

### DIFF
--- a/language/en/common.php
+++ b/language/en/common.php
@@ -58,6 +58,6 @@ $lang = array_merge($lang, array(
 	'AD_POP_UP_DESC'				=> 'Displays once per day when user visits this board as overlaying box. User need to close this box to continue to the content. Please, be aware, that this kind of advertisement is very obtrusive to the user!',
 	'AD_SLIDE_UP'					=> 'Slide up',
 	'AD_SLIDE_UP_DESC'				=> 'Displays on every page after user scrolls below main content. Slides up from the bottom.',
-	'AD_HIDDEN'						=> 'Hidden',
-	'AD_HIDDEN_DESC'				=> 'Is present on every page, but never visible. This location can be used for tracking codes, automatic ads, etc.',
+	'AD_SCRIPTS'					=> 'Scripts',
+	'AD_SCRIPTS_DESC'				=> 'This location is for specialty JavaScript code like AdSense Auto ads, tracking codes, etc. The code entered here will be inserted into the page’s HEAD tag and is not intended for ad placement, but only for helper scripts.',
 ));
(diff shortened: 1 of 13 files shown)


language/en/common.php
```diff
@@ -58,4 +58,6 @@
 	'AD_POP_UP_DESC'				=> 'Displays once per day when user visits this board as overlaying box. User need to close this box to continue to the content. Please, be aware, that this kind of advertisement is very obtrusive to the user!',
 	'AD_SLIDE_UP'					=> 'Slide up',
 	'AD_SLIDE_UP_DESC'				=> 'Displays on every page after user scrolls below main content. Slides up from the bottom.',
+	'AD_HIDDEN'						=> 'Hidden',
+	'AD_HIDDEN_DESC'				=> 'Is present on every page, but never visible. This location can be used for tracking codes, automatic ads, etc.',
```
VSEphpbb: instead of `automatic ads` call it `AdSense Auto ads`

language/en/common.php
```diff
@@ -58,4 +58,6 @@
 	'AD_POP_UP_DESC'				=> 'Displays once per day when user visits this board as overlaying box. User need to close this box to continue to the content. Please, be aware, that this kind of advertisement is very obtrusive to the user!',
 	'AD_SLIDE_UP'					=> 'Slide up',
 	'AD_SLIDE_UP_DESC'				=> 'Displays on every page after user scrolls below main content. Slides up from the bottom.',
+	'AD_HIDDEN'						=> 'Hidden',
```
VSEphpbb: I'm not sure it's best to call this "Hidden" and to make it hidden HTML.

Google says Auto Ads should go inside the `HEAD` tag, so lets do that (using `overall_header_stylesheets_after`).

Let's instead of calling this `Hidden`, call it `Scripts`, and describe this as a location where users can ad javascript codes like AdSense Auto ads, tracking codes, etc. to the HEAD element of their forum's pages. We should make it clear this is not for ad placement, but only for *helper* scripts.

Lets also make AdSense Auto ads in the description a hyperlink so users will be able to read about it and hopefully fully understand what it is and how it works and how to use it, https://adsense.googleblog.com/2018/02/introducing-adsense-auto-ads.html

So the new description below could be like:

This location is for specialty JavaScript code like <a href="https://adsense.googleblog.com/2018/02/introducing-adsense-auto-ads.html">AdSense Auto ads</a>, tracking codes, etc. The code entered here will be inserted into the page's HEAD tag and is not intended for ad placement, but only for helper scripts.

senky: Desc is only displayed as a title of the `<option>` so it won't be a link. Other than that, I agree with your renaming and have done it.

location/type/scripts.php
```diff
@@ -0,0 +1,30 @@
+<?php
+/**
+ *
+ * Advertisement management. An extension for the phpBB Forum Software package.
+ *
+ * @copyright (c) 2017 phpBB Limited <https://www.phpbb.com>
```
VSEphpbb: We should make this 2018

PR #136, opened 2018-12-19
Fix get_ads() with empty result from DB

### LINKED ISSUE #135: Please fix
I found errors

phpBB Debug] PHP Warning: in file [ROOT]/ext/phpbb/ads/ad/manager.php on line 105: array_filter() expects parameter 1 to be array, boolean given
[phpBB Debug] PHP Warning: in file [ROOT]/ext/phpbb/ads/event/main_listener.php on line 133: Invalid argument supplied for foreach()

senky:
Fixes #135

### DIFF
--- a/ad/manager.php
+++ b/ad/manager.php
@@ -97,6 +97,11 @@ class manager
 		$data = $this->db->sql_fetchrowset($result);
 		$this->db->sql_freeresult($result);
 
+		if (empty($data))
+		{
+			return [];
+		}
+
 		$current_location_id = '';
 		$data = array_filter($data, function ($row) use (&$current_location_id) {
 			$return = $current_location_id !== $row['location_id'];


ad/manager.php
```diff
@@ -97,6 +97,11 @@ public function get_ads($ad_locations, $user_groups, $non_content_page = false)
 		$data = $this->db->sql_fetchrowset($result);
 		$this->db->sql_freeresult($result);
 
+		if (empty($data))
```
VSEphpbb: I think the real issue here is, why would the query on line 96 error, thuse returning the false boolean that is what eventually gets sent to array_filter below, causing the error???

senky: Because when there is no result, $data from line 97 === false; thus `array_filter` complains that it's first parameter is bool not an array. But since it spans across multiple lines, error code reports the last line of the function.

VSEphpbb: But when there are no ads, there is still a $result object:

```
mysqli_result Object
(
    [current_field] => 0
    [field_count] => 4
    [lengths] => 
    [num_rows] => 0
    [type] => 0
)
```
and so `$data` is correctly returned as an empty array by `sql_fetchrowset`. 

screen shot 2018-12-18 at 11 33 32 pm

The only way this all goes wrong is if the original query fails.

I have no ads and am not having any errors.

I think the reporter of the issue needs to provide more information if possible since this has never been reported before. My user-error alarm bells are ringing 😉

fonziset:
it does so in expanding

https://github.com/PGreca/pgsocial

PR #138, opened 2019-03-01
Small updates

VSEphpbb:


controller/helper.php
```diff
@@ -107,10 +107,10 @@ public function assign_data($data, $errors)
 	/**
 	 * Assign template locations data to the template.
 	 *
-	 * @param	mixed	$ad_locations	The form data or nothing.
-	 * @return	void
+	 * @param array $ad_locations The form data or nothing.
+	 * @return void
 	 */
-	public function assign_locations($ad_locations = false)
+	public function assign_locations($ad_locations = array())
```
senky: You sure this won't cause any problems? I don't remember why we used false, but it probably has something to do with DB query which returns false in case there are no rows I assume.

VSEphpbb: Should be fine. Everywhere `ad_locations` is assigned, it's always an array.

PR #152, opened 2020-04-22
Escape last in filter chains

### LINKED ISSUE #151: Broken popup header title output
https://github.com/phpbb-extensions/ad-management/blob/dcbe246fee2cb7a0aa94e33112f91fb3a31d90dd/styles/all/template/includes/phpbb_ads_pop_up.html#L7

```twig
{{ lang('ADVERTISEMENT')|e('js')|upper }}
```
Output for `en`: `Advertisement` -> `ADVERTISEMENT`
Output for `ru`:  `Реклама` -> `\U0420\U0435\U043A\U043B\U0430\U043C\U0430`

`e('js')` does that. Either Twig bug or some argument/setting to escape non-latin chars is missing.

P.S. Or probably this is `upper` which makes valid UTF8 chars invalid ( valid`\u0420` -> invalid (`\U0420`). Possibly `upper` is redundant here as css does it all.

VSEphpbb:
Fixes #151


rxu:
That fixed it, thanks.

PR #185, opened 2024-12-25
Use storage in ads uploads

rubencm:


config/routing.yml
```diff
@@ -15,3 +15,8 @@ phpbb_ads_visual_demo:
     defaults: { _controller: phpbb.ads.visual_demo.controller:handle }
     requirements:
         action: enable|disable
+
+phpbb_ads_storage_banner:
```
iMattPro, line 19: Where is this being used, or why is this new controller needed?

migrations/v30x/m1_storage.php
```diff
@@ -0,0 +1,43 @@
...
+			'\phpbb\db\migration\data\v400\dev',
+			'\phpbb\ads\migrations\v20x\m1_hide_ad_for_group',
+		);
+	}
+
+	public function update_data()
+	{
+		return array(
+			['config.add', ['storage\\phpbb_ads\\provider', local::class]],
+			['config.add', ['storage\\phpbb_ads\\config\\path', 'images/phpbb_ads']], // todo: make sure this exists in migration if is a new installation
```
iMattPro, line 32: This _to do_ should be easy
```php
public function effectively_installed()
{
	return $this->config->offsetExists('storage\\ phpbb_ads\\config\\path');
}
```

rubencm:
Another todo: Old files need to be tracked
https://github.com/phpbb/phpbb/blob/ad135f3f1b793198ac146870657df31e940ef2df/phpBB/phpbb/db/migration/data/v400/storage_track.php
https://github.com/phpbb/phpbb/blob/ad135f3f1b793198ac146870657df31e940ef2df/phpBB/phpbb/db/migration/data/v400/storage_backup_data.php

migrations/v30x/m1_storage.php
rubencm: I mean, make sure that the folder exist

config/routing.yml
rubencm: To load the banner images through the controller, since now they can be stored anywhere.
If the storage is not local, they won't be in images/phbbb_ads

iMattPro: Ahh, OK

rubencm: There is a couple example extensions in my profile if you want to check
(but they work only with my pr "improve storage")

config/services.yml
```diff
@@ -4,6 +4,24 @@ imports:
     - { resource: location.yml }
 
 services:
+    phpbb.ads.storage:
+        class: phpbb\storage\storage
+        arguments:
+            - '@storage.adapter.factory'
+            - '@storage.file_tracker'
```
iMattPro, line 11: `@storage.file_tracker` comes up as a non-existent service

rubencm: You need to use this version https://github.com/phpbb/phpbb/pull/6753

migrations/v30x/m1_storage.php
iMattPro: Oh ok, yeah that folder was never created by this extension until someone tried to upload something for the first time.

migrations/v30x/m1_storage.php
```diff
@@ -0,0 +1,97 @@
...
+		if ($handle)
+		{
+			while (($file = readdir($handle)) !== false)
+			{
+				if ($file === '.' || $file === '..')
+				{
+					continue;
+				}
+
+				$file_tracker->track_file('phpbb_ads', $file, filesize($dir . '/' . $file));
```
iMattPro, line 77: This makes me wonder if the `$file_tracker` class should have `track_files` and `untrack_files` methods for handling multiple entries in 1 query?

rubencm: Yes, it can be updated to support more than one file.
I add it to the todo list

iMattPro: Great! (Will be a lot less queries)

config/routing.yml
```diff
@@ -15,3 +15,8 @@ phpbb_ads_visual_demo:
...
+    path:             /ads_download/{file}
+    defaults:
+        _controller:  phpbb.ads.controller.banner:handle
```
iMattPro, line 22: We'll probably want to add `stateless: true` here eventually, re: https://github.com/phpbb/phpbb/compare/master...marc1706:phpbb:feature/stateless_sessions?expand=1

rubencm: Well, this can be done in the future, there is not even a pull request for this and can change in the future

